PR #1, opened 2016-05-21
Now with 200% more Coveralls

ararslan:
I added the coverage stuff to .travis.yml and added the coverage badge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,5 @@ notifications:
 #script:
 #  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
 #  - julia -e 'Pkg.clone(pwd()); Pkg.build("Primes"); Pkg.test("Primes"; coverage=true)'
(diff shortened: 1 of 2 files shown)


tkelman:
@ararslan would you be up for having appveyor run under your name here? You can log in via github auth, if it doesn't let you enable the repo directly then we can go through a couple manual steps to do so.

tkelman:
It's not very easy to run CI tests on 32 bit Linux at the moment, but it is easy to run AppVeyor with the 32 bit Windows binaries of Julia (the VM's are 64 bit, but that doesn't make much difference). With the large numbers this package involves it will be useful to keep testing against 32 bit to avoid overflowing there.

ararslan:
@tkelman See #2. I added you as a collaborator on my AppVeyor account so hopefully that means you can trigger builds. I have only a vague idea of how to use AppVeyor... 😕

Issue #3, opened 2016-05-22
Docs


tkelman:
The docstrings for these aren't terribly long, so maybe the easiest thing to do is just copy them into the readme. There's also Documenter.jl but that may be overkill here.

ararslan:
I took a stab at it. Let me know what you think.

ararslan:
Addressed in #4.

PR #4, opened 2016-05-22
Copied docstrings to README

ararslan:
Addresses #3. I'm not crazy about the formatting; I tried to mimic the look of the docs and I don't know how well it translated to markdown.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,45 @@
 Julia functions for computing prime numbers.
 
 At the moment, this repository just contains the following functions which have been duplicated from Base Julia:
-* `isprime`
-* `primes`
-* `primesmask`
-* `factor`
+
+    factor(n) -> Dict
+
+> Compute the prime factorization of an integer `n`. Returns a dictionary. The keys of the
+dictionary correspond to the factors, and hence are of the same type as `n`. The value
+associated with each key indicates the number of times the factor appears in the
+factorization.
+> ```julia
+julia> factor(100) # == 2*2*5*5
+Dict{Int64,Int64} with 2 entries:
+  2 => 2
+  5 => 2
+> ```
+
+    isprime(x::Integer) -> Bool
+
+> Returns `true` if `x` is prime, and `false` otherwise.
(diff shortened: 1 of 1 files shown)


ararslan:
Hm, having decreased coverage show as a failed check is a little unfortunate. @tkelman Any idea if there's a better way to set up the coverage testing? I don't think Julia has it as a check like this.

tkelman:
This looks fine to me, thanks! This was the last thing needed before registering as far as I'm concerned.

There might be a way to configure coveralls to allow small coverage drops, otherwise I think most people just ignore it.

ararslan:
@tkelman Available PR alert settings for Coveralls are:
* Leave comments? (checked)
* Use status API? (checked)
* Coverage threshold for failure (blank)
* Coverage decrease threshold for failure (blank)

I don't know what any of those mean. I'd personally prefer not to have decreased coverage show up as a failed check because it makes it less obvious whether Travis and AppVeyor failed.

tkelman:
I think what you want is to disable "Use status API" then, that way it won't submit a data point for the checkbox on PR's. I personally find the PR comments annoying noise, but blocking Coveralls and Codecov's bots at least made my inbox a little quieter.

ararslan:
I can turn those off too if you'd prefer.

tkelman:
Some people like them, and if you'd rather not have coverage info show up in the status API then they won't otherwise be visible in PR's right away. I think setting one of the thresholds for failure (dunno what the difference is) would have coverage info available but not flag a failure?

ararslan:
I think the first listed threshold for failure is a percent coverage below which any check would fail. The second threshold is the allowable percent decrease in coverage, so if you submit a PR that decreases coverage by x+1% and the threshold is x% then it fails. That's what I'd guess anyway.

The latter could be a good idea, it's just a matter of deciding on a meaningful threshold. For example, copying the docs in the README decreased coverage by 2% (which I honestly don't understand at all) so clearly 2% isn't a good threshold...

PR #12, opened 2016-05-29
WIP: Incorporate history from base

rfourquet:
This is a follow-up of https://github.com/JuliaLang/julia/issues/16357#issuecomment-222144998 (cc @tkelman). I extracted from base the history of primes.jl. I thougth it was not necessary to bother with older history (which was in intfuncs.jl). I also didn't include changes to test/numbers.jl which have tests for primes functions (I don't know how to include only the parts relevant to primes). I see those possibilities (besides doing nothing!):
1. Rebase the history of the package on this branch and make a PR of the result
2. Merge as is (by first deleting base/primes.jl)
3. Someone simply push this branch to the repo, unmerged, as a reference. It would simply be a convenience when hacking on Primes.jl, to avoid having to check the history in the julia repo.

I think I prefer option 3. 
For reference, this is how I extracted the history:
```bash
function cherry() {
    while read c; do
        git cherry-pick --strategy=recursive -X theirs $c;
        if [ $? != 0 ]; then
            # could add: '! -wholename ./test/numbers.jl '
            find . -path ./.git -prune -o -type f ! -wholename ./base/primes.jl -exec rm {} +
            git add -u
            git commit --no-edit
        else
            find . -path ./.git -prune -o -type f ! -wholename ./base/primes.jl -exec rm {} +
            git add -u
            git commit --amend --no-edit
        fi
    done;
}

function rewriteprimes () {
    git reset --hard a9cbc036ac62dc5ba5200416ca7b40a2f9aa59ea # initial commit
    # last commit listed by 'git rev-list' is the removal of primes.jl
    git rev-list  --reverse master -- base/primes.jl | head -n 33 | cherry
}
```
What I wanted was to pick the file from base as is and the corresponding commit, I hope cherry-pick resulted in something equivalent.
I had to artificially rebase the result on the initial commit of the package to be able to open a PR here.


rfourquet:
After moving it's not directly mergeable, nor is it possible to rebase master onto this branch (option 1 above) without having conflicts to solve (in particular in master the src/Primes.jl file is created almost empty, containing the module declaration and comments). On the other hand, if I then 'git checkout master -- src/Primes.jl' and commit, we get a small diff (essentially docstrings added inline and a couple of small changes), I updated the branch accordingly.
Note that by option 1 I don't mean modifying this package history, but rather to create a parallel one based on master's one, which can then be merged in master.

src/Primes.jl
```diff
@@ -1,4 +1,13 @@
-# This file is a part of Julia. License is MIT: http://julialang.org/license
```
tkelman: I think we should keep this, just replace "is a part" with "was formerly a part"

PR #13, opened 2016-06-02
RFC: Added a function to list all prime factors with multiplicity

ararslan:
Sometimes it's nice to be able to have an array with *all* of the prime factors of a number. This PR introduces a convenience function `factorize` that does just that.
```julia
julia> factorize(4)
2-element Array{Int64,1}:
 2
 2
```
This reshapes the output from
```julia
julia> factor(4)
Dict{Int64,Int64} with 1 entry:
  2 => 2
```
Since it's essentially collecting a dictionary, there's *no guaranteed order* to the output array, other than the fact that the runs of primes are grouped. I don't know whether that's an issue.

Does this seem like a useful function to include in the package?

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -209,7 +209,7 @@ Dict{Int64,Int64} with 2 entries:
 ```
 """
 function factor{T<:Integer}(n::T)
-    0 < n || throw(ArgumentError("number to be factored must be > 0, got $n"))
+    0 < n || throw(ArgumentError("number to be factored must be ≥ 0, got $n"))
     h = Dict{T,Int}()
     n == 1 && return h
     isprime(n) && (h[n] = 1; return h)
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -108,7 +108,7 @@ function primes(lo::Int, hi::Int)
     lo ≤ 3 ≤ hi && push!(list, 3)
     lo ≤ 5 ≤ hi && push!(list, 5)
     hi < 7 && return list
-    sizehint!(list,  5 + floor(Int, hi / (log(hi) - 1.12) -  lo / (log(max(lo,2)) - 1.12*(lo > 7))) ) # http://projecteuclid.org/euclid.rmjm/1181070157    
+    sizehint!(list,  5 + floor(Int, hi / (log(hi) - 1.12) -  lo / (log(max(lo,2)) - 1.12*(lo > 7))) ) # http://projecteuclid.org/euclid.rmjm/1181070157
     sieve = _primesmask(max(7, lo), hi)
     lwi = wheel_index(lo - 1)
     @inbounds for i = 1:length(sieve)   # don't use eachindex here
@@ -209,7 +209,7 @@ Dict{Int64,Int64} with 2 entries:
 ```
 """
 function factor{T<:Integer}(n::T)
-    0 < n || throw(ArgumentError("number to be factored must be ≥ 0, got $n"))
+    0 < n || throw(ArgumentError("number to be factored must be > 0, got $n"))
     h = Dict{T,Int}()
     n == 1 && return h
     isprime(n) && (h[n] = 1; return h)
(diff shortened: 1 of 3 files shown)


src/Primes.jl
```diff
@@ -108,7 +108,7 @@ function primes(lo::Int, hi::Int)
     lo ≤ 3 ≤ hi && push!(list, 3)
     lo ≤ 5 ≤ hi && push!(list, 5)
     hi < 7 && return list
-    sizehint!(list,  5 + floor(Int, hi / (log(hi) - 1.12) -  lo / (log(max(lo,2)) - 1.12*(lo > 7))) ) # http://projecteuclid.org/euclid.rmjm/1181070157
```
ararslan: Looks like my editor auto-ate some trailing whitespace

tkelman:
trailing whitespace serves no purpose anyway.

there's a bit of an unfortunate naming collision with the matrix factorize function. guess it's okay if not exported here, but potentially confusing.

ararslan:
Wow, hard facepalm for me. I forgot there already is a `factorize` function in base. Any other name suggestions? Initially I had considered `factors`, but that seems confusingly close to `factor`.

simonbyrne:
Yeah, I think the name might be a bit too general (I think `factor` is probably too general as well, if we were doing this from scratch I would have suggested `primefactor`).

Perhaps `factorvec` might be more descriptive?

ararslan:
Hm, `factorvec` isn't bad but it doesn't seem immediately descriptive to me. I'd be okay with it though if folks like it. I can't think of a suitable alternative at the moment anyway.

We could always channel our inner Java and go the super verbose route and use `primefactorization`, or even better, `primefactorswithmultiplicity`. 😜

ararslan:
Hm, you know @simonbyrne, this *is* a separate package, so perhaps we could get away with renaming `factor` to `primefactor`, or at least deprecating `factor` in favor of `primefactor`. If we were to do that, `primefactorization` for this PR would be a natural analog, though it still seems annoyingly long to me.

ararslan:
More on the topic of this particular PR, I'd love to hear what people think about the fact that this function, whatever its name may be, will return a vector with no guaranteed order. Would it make sense to `sort` it first, or is it okay to simply state the caveat?

simonbyrne:
Hopefully #14 will help with the sorting.

It occurs to me that you're treating the output of `factor` as a [run-length encoding](https://en.wikipedia.org/wiki/Run-length_encoding), and expanding out the result.

ararslan:
Yep, that's exactly what I'm after.

ararslan:
Alright, now that #14 has been merged, all of the sorting stuff for this PR should be a nonissue. I'll resolve the branch conflicts later today. `factorvec` is a good enough name as far as I'm concerned so in the absence of any opposition I'll update to that.

ararslan:
Full disclosure: I don't know what I'm doing. No idea how to resolve conflicts. 😕

ararslan:
Uhhhhhhh *WHOOPS*. I did **not** mean to do that.

ararslan:
> Pull request successfully merged and closed

It wasn't merged though?? And I didn't want it to be, I just wanted to resolve conflicts. Help please 😨

ararslan:
Okay cool, I was able to clean up the huge mess I made. That was an adventure.

Since this adds functionality that's never been in base, unlike the other functions, should it be exported regardless of the version? If so I can it put in a separate `export` outside of the conditional.

rfourquet:
I think `factorvec` is not a good enough descriptive name, does "vec" stands for the output type? I unfortunately have no other suggestions for now... The name `multifactors` comes to my mind ("multi" as in "multiset"), but then it's more a name whereas `factor` is a verb, this may be okay though.

rfourquet:
I wonder if this functionality is really needed enough (in general) to warrant its inclusion here. I never saw a dedicated function for that in the few math packages I know, in particular in pari which is dedicated to number theory. And it's also not difficult to implement for those really needing it.

ararslan:
@rfourquet This is just a convenience function; it's not particularly complex but I don't think it would hurt to have it in here. Regarding the name, I assume Simon's reasoning for `vec` was indeed that it returns a vector. While I'm not super crazy about `factorvec`, I think it is fairly descriptive: `factor` (get all prime factors, as with the `factor` function) `vec` (return as a vector).  ¯\\\_(ツ)\_/¯

rfourquet:
Maybe we won't be able to come up with a better name, but it's more difficulut to change it now that it's merged. It would have be nice to wait a bit and get more feedback. And: do we want this function to be exported? I also suggest that you squash the commits next time.

ararslan:
I meant to squash on merge but forgot this time.

ararslan:
`factorvec` is a good enough name. For consistency it'll be exported when the rest are, in 0.6-pre, and unexported in the meantime.

PR #14, opened 2016-06-05
use a SortedDict for factor

simonbyrne:
The main advantage is for iteration, so that the factors are returned in increasing order.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.4
+DataStructures
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -1,6 +1,8 @@
 __precompile__()
 module Primes
 
+import DataStructures: SortedDict
+
 if VERSION >= v"0.5.0-dev+4340"
     if isdefined(Base,:isprime)
         import Base: isprime, primes, primesmask, factor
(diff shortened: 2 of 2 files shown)


ararslan:
The introduction of a dependency isn't ideal but probably worth it in this case.

rfourquet:
While I appreciate to have a sorted order at the terminal, it's not clear for me what is specific to prime factorization which makes a `SortedDict` objectively better than the default? It's easy enough to construct a `SortedDict` from a `Dict`, so I think it's not worth the dependency on `DataStructure`.

ararslan:
@rfourquet It makes the order of the output in #13 predictable.

rfourquet:
@ararslan I'm not convinced, as it is trivial to call sort in the function in #13 to achieve predictability.

ararslan:
Sure, but #13 aside, as Simon said in his original comment, it provides an advantage for iteration. For example, it guarantees that `keys(factor(n))` will be sorted.

rfourquet:
Yes, `SortedDict(factor(n)))` also guarantees that. But again what is specific to prime factorization? (I guess my point is that if having sorted keys was always an advantage, then it's up to Base to switch its implementation to a SortedDict). Again I agree it's nice to have a sorted list of factors, but some users will prefer to have them in the default associative data-structure from Base, and they will now have to call `Dict(factor(n))` to get the desired behavior. So the change in this PR is not a win-win.

PR #16, opened 2016-06-07
add names to license

simonbyrne:
I just took the names who had made more than 1 commit.


tkelman:
should also put back the header from base

simonbyrne:
good idea. done

ararslan:
I gotta say, I'm honored to see my name in that list :D

PR #18, opened 2016-06-09
speed up isprime by fusing the two loops on small primes

rfourquet:
This change consistently speeds up isprime on my machine, e.g. by about 70% for random numbers less than 2^32, and by 15% for numbers less than 2^64.
cc @pabloferz

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -137,10 +137,8 @@ function isprime(n::Integer)
     # Small precomputed primes + Miller-Rabin for primality testing:
     #     https://en.wikipedia.org/wiki/Miller–Rabin_primality_test
     #     https://github.com/JuliaLang/julia/issues/11594
-    iseven(n) && return n == 2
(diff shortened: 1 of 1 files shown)


src/Primes.jl
```diff
@@ -137,9 +137,8 @@ function isprime(n::Integer)
     # Small precomputed primes + Miller-Rabin for primality testing:
     #     https://en.wikipedia.org/wiki/Miller–Rabin_primality_test
     #     https://github.com/JuliaLang/julia/issues/11594
-    iseven(n) && return n == 2
-    n in (3,5,7,11,13,17,19,23) && return true
-    for m in (3,5,7,11,13,17,19,23)
+    for m in (2,3,5,7,11,13,17,19,23)
+        n == m && return true
         n % m == 0 && return false
```
pabloferz: Maybe even `n % m == 0 && return m == n`?

rfourquet: Yes good idea indeed, updated accordingly.

PR #19, opened 2016-06-09
choose the container holding the factors via an argument

rfourquet:
This is an alternative API to #13, and a complement to #14. The user can choose to have `factor` output a `Dict`, a `SortedDict` or a `Vector` via an argument: `factor(12, Dict)`, `factor(12, SortedDict)`, `factor(12, Vector)` (calls `factorvec`, which would then be not exported). The default of `factor(12)` is `Dict` for now, but this will obviously be discussed! (if the idea gets support).

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -2,13 +2,11 @@
 __precompile__()
 module Primes
 
-import DataStructures: SortedDict
-
 if VERSION >= v"0.5.0-dev+4340"
     if isdefined(Base,:isprime)
         import Base: isprime, primes, primesmask, factor
     else
-        export isprime, primes, primesmask, factor, factorvec
+        export isprime, primes, primesmask, factor
     end
 end
 
@@ -221,23 +219,22 @@ isprime(n::Int128) = n < 2 ? false :
 #     http://maths-people.anu.edu.au/~brent/pub/pub051.html
 #
 """
-    factor(n) -> DataStructures.SortedDict
+    factor(n) -> Dict
 
-Compute the prime factorization of an integer `n`. Returns a sorted dictionary. The
+Compute the prime factorization of an integer `n`. Returns a dictionary. The
 keys of the dictionary correspond to the factors, and hence are of the same type as `n`.
 The value associated with each key indicates the number of times the factor appears in the
 factorization.
 
 ```jldoctest
 julia> factor(100) # == 2*2*5*5
-DataStructures.SortedDict{Int64,Int64,Base.Order.ForwardOrdering} with 2 entries:
+Dict{Int64,Int64} with 2 entries:
   2 => 2
   5 => 2
 ```
 """
-function factor{T<:Integer}(n::T)
+function factor{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int}=Dict{T,Int}())
     0 < n || throw(ArgumentError("number to be factored must be > 0, got $n"))
-    h = SortedDict{T,Int,Base.Order.ForwardOrdering}()
     n == 1 && return h
     isprime(n) && (h[n] = 1; return h)
     local p::T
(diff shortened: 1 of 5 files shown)


src/Primes.jl
```diff
@@ -258,6 +257,11 @@ function factor{T<:Integer}(n::T)
     T <: BigInt || widemul(n-1,n-1) <= typemax(n) ? pollardfactors!(n, h) : pollardfactors!(widen(n), h)
 end
 
+factor{T<:Integer}(n::T) = factor(n, Dict{T,Int}())
+factor{T<:Integer}(n::T, ::Type{SortedDict}) =
```
ararslan: If I'm not mistaken I think the convention is for type specifications to come before the other arguments, i.e. `factor(::Type{SortedDict}, n::T)`.

simonbyrne:
+1: This seems like a great solution, though I agree with @ararslan that the output type should be the first argument.

tkelman:
or a keyword argument, which has precedent in JSON.jl

tkelman:
this would also make the DataStructures requirement not needed - it could be moved to test only perhaps if we have specific things to test about using a SortedDict

rfourquet:
I hesitated for the order of arguments, I agree it should be changed. I also thought about a keyword, but dismissed this solution for the type instability. This may not be problematic though.
I updated by removing the dependency on DataStructures, by exploiting the fact that `SortedDict(d)` just works (where `d` is a `Dict`). Compared to the last version, this is slightly suboptimal as we end up constructing first `d` then its sorted version, but it should be a negligible overhead.

src/Primes.jl
```diff
@@ -2,8 +2,6 @@
 __precompile__()
 module Primes
 
-import DataStructures: SortedDict
```
tkelman: require as well, if this was the last mention?

rfourquet: last mention == final version? Good catch, I may have forgotten about REQUIRE. I will wait for feedback on the API before finalizing.

tkelman: last mention meaning does it appear anywhere else in the code base? could be answered by applying the patch then git grep, but I don't think I have a clone handy

simonbyrne: Maybe include it in the test suite though, in which case you should add it to the `tests/REQUIRE` file.

rfourquet: OK, I made a tests/REQUIRE file, and the only "using DataStructures" mention is in runtests.jl. But a "jldoctest" in Primes.jl still contains a reference to `SortedDict`, is it a problem?

tkelman: good question. do we have doctest running hooked up here? if not, can wait to figure that one out

test/runtests.jl
```diff
@@ -209,3 +210,7 @@ euler7(n) = primes(floor(Int,n*log(n*log(n))))[n]
 @test factorvec(1) == Int[]
 @test factorvec(3) == [3]
 @test factorvec(4) == [2,2]
+
+# factor which non-default containers
```
tkelman: with

src/Primes.jl
ararslan: @tkelman I don't believe so

ararslan:
This is a great way to do it, thanks for this. We're back to `factor(Vector, n)` not having a predictable order, though perhaps a `sort` wouldn't add too much overhead since the number of distinct elements will typically be small. We could do away with `factorvec` entirely and just move the source to
```julia
factor{T<:Integer}(::Type{Vector}, n::T) = ...
```
and wrap the `mapreduce` in a `sort`.

rfourquet:
The `sort!` is already added! Indeed we could do away with `vectorvec`, will change according to your suggestion.

ararslan:
Oh nice! I missed the `sort!` reading through the source.

rfourquet:
`vectorvec` has been short-lived!

src/Primes.jl
```diff
@@ -221,23 +219,22 @@ isprime(n::Int128) = n < 2 ? false :
...
 ```jldoctest
 julia> factor(100) # == 2*2*5*5
-DataStructures.SortedDict{Int64,Int64,Base.Order.ForwardOrdering} with 2 entries:
+Dict{Int64,Int64} with 2 entries:
   2 => 2
   5 => 2
 ```
 """
-function factor{T<:Integer}(n::T)
+function factor{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int}=Dict{T,Int}())
```
ararslan: Would it make more sense to leave the basic API as `factor{T<:Integer}(n::T)` and just have `factor{T:<Integer, D<:Associative}(::Type{D}, n::T) = D(factor(n))`?

rfourquet:
Rebased (hadn't noticed a merge conflict caused by #20, which I don't understand as the touched lines don't overlap). 
Anyone has some reservations with this change?

src/Primes.jl
ararslan: So if I'm understanding you correctly, if a dictionary is passed that already contains values, the factorization of the integer will be added to the dictionary? I'm kind of confused on how that works. Would you mind adding a test to verify the behavior you expect there?

ararslan:
The AppVeyor failure appears to be both unrelated and strange.

src/Primes.jl
rfourquet: Yes, it's what I mean. My idea is that we should have multiple routines with this API: takes as input the number `n` to factor and a dict containing potentially already found factors by other routines, and outputs the same dict updated with newly found factors, and the divisor of `n` which remains to be factored. I have a need for that in a package I am making now.

If you don't mind, I would rather put up (can be very soon) a new PR with my proposed API update, and include tests there. My change here to pass a `Dict` to the `factor` method was just a tiny step in this direction, but remains an implementation detail which is not part yet of a documented API.

ararslan: Sure, that's fine with me. I'm looking forward to seeing your idea manifest in another PR. :)

ararslan:
I have no objections and no one else has spoken up so I'd say we're good to go. Great work here, @rfourquet!

rfourquet:
Thanks! But really my "idea" is modest. Maybe I will put the above mentioned package online before opening the PR, to show a real use case.
@tkelman: would you "invite" to this repo?

PR #21, opened 2016-06-17
Add a Set method for factor, plus a specific error fallback

ararslan:
`factor(Set, n)` makes a set of the distinct prime factors. Calling `factor` with a type argument that's anything other than the ones documented is a specific `MethodError`. Currently it's a weird leftover error from Base that says to use the Primes package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Dict{Int64,Int64} with 2 entries:
     factor(ContainerType, n) -> ContainerType
 
 > Return the factorization of `n` stored in a `ContainerType`, which must be a
-subtype of `Associative`, `AbstractArray`, or `Base.AbstractSet`.
+subtype of `Associative` or `AbstractArray`, a `Set`, or an `IntSet`.
 
 > ```julia
 julia> factor(DataStructures.SortedDict, 100)
@@ -47,16 +47,18 @@ julia> factor(Vector, 100)
  2
  5
  5
+
+julia> prod(factor(Vector, 100)) == 100
+true
 > ```
 
-> When `ContainerType <: Base.AbstractSet`, this returns the distinct prime
+> When `ContainerType == Set`, this returns the distinct prime
 factors as a set.
 
 > ```julia
 julia> factor(Set, 100)
 Set([2,5])
 > ```
-
 
     isprime(x::Integer) -> Bool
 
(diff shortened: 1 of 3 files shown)


rfourquet:
Makes sense. I had thought that `keys(factor(n))` would be enough to get unique factors, but there is no reason to leave `Set` out of the API.

ararslan:
I assume `keys` would still be the typical way to get the unique factors, `Set` is just convenient sometimes.

rfourquet:
Oh you just changed the type constraints from `Base.AbstractSet` to `Union{Set,IntSet}`? was there a problem with the former? Would be nice to support `DataStructures.SortedSet`.

ararslan:
`Base.AbstractSet` isn't documented or exported, so I assume it shouldn't be used. Also it doesn't exist in 0.4 (which I didn't realize before submitting the PR 😕). Does `DataStructures.SortedSet` inherit from `Base.AbstractSet` in 0.5?

rfourquet:
Oh I assumed so, but is doesn't... the only mention of `AbstractSet` I can find in the repo is "TODO: Most of these functions should be removed once AbstractSet is introduced there" (https://github.com/JuliaLang/DataStructures.jl/search?utf8=%E2%9C%93&q=AbstractSet). So it may be a simple matter of opening a PR. 
The fact that `AbstractSet` is not exported in Base seems to be an oversight, there was an issue (#5533) and a PR (#12816) for its introduction. So I would say if it complicates 0.4 compatibility, keep as is for now, otherwise use `AbstractSet`.

ararslan:
Yeah, it currently wrecks 0.4 so this should probably stay as-is for now, but this is definitely something to keep an eye on and update going forward. Thanks for the links and research!

I left a comment on Jameson's PR that introduced `AbstractSet` in Base asking whether it should be exported. If that gets exported we can get in touch with DataStructures.jl about it.

rfourquet:
Ok looks good to go! but let's give some time for others to comment (and then for a rebase).
Your test that `factor(Tuple, n)` errors makes me wonder if it should be supported, but this question can be postponed until a need arises.

ararslan:
No one else has spoken up and I want to get this error fix in there so I'm going to go ahead and merge this. It's pretty non-disruptive anyway.

PR #22, opened 2016-06-17
allow factoring non-positive numbers

rfourquet:
This adds convenience extensions to "pure" factorization of natural numbers. The first change seems to be rather uncontroversial among computer algebra systems (e.g. Magma, SageMath, Pari): a negative number has -1 as a factor:
```
julia> factor(-9)
Dict{Int64,Int64} with 2 entries:
  -1 => 1
   3 => 2
```
The second change concerns zero, wich is factored as `0^1`. This erros on Magma and SageMath, but is defined in Pari, which was built by world-class number theorist (its creator has been my teacher in master :-p), so it would be credible to do the same. 
```
julia> factor(0)
Dict{Int64,Int64} with 0 entries:
  0 => 1
```

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -8,7 +8,18 @@ if VERSION >= v"0.5.0-dev+4340"
     else
         export isprime, primes, primesmask, factor
     end
+    using  Base: BitSigned
+    using Base.Checked.checked_neg
+else
+    typealias BitSigned Union{Int128,Int16,Int32,Int64,Int8}
+    function checked_neg(x::Integer)
+        y = -x
+        (y<0) == (x<0) && throw(OverflowError())
+        y
+    end
 end
+
+
 
 # Primes generating functions
 #     https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes
@@ -219,7 +230,7 @@ isprime(n::Int128) = n < 2 ? false :
 #     http://maths-people.anu.edu.au/~brent/pub/pub051.html
 #
 """
-    factor(n) -> Dict
+    factor(n::Integer) -> Dict
 
 Compute the prime factorization of an integer `n`. Returns a dictionary. The
 keys of the dictionary correspond to the factors, and hence are of the same type as `n`.
@@ -232,11 +243,41 @@ Dict{Int64,Int64} with 2 entries:
   2 => 2
   5 => 2
 ```
+
+For convenience, a negative number `n` is factored as `-1*(-n)` (i.e. `-1` is considered
+to be a factor), and `0` is factored as `0^1`:
+
+```jldoctest
+julia> factor(-9) # == -1*3^2
+Dict{Int64,Int64} with 2 entries:
+  -1 => 1
+   3 => 2
+
+julia> factor(0)
+Dict{Int64,Int64} with 1 entries:
+  0 => 1
+```
 """
 function factor{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int}=Dict{T,Int}())
-    0 < n || throw(ArgumentError("number to be factored must be > 0, got $n"))
-    n == 1 && return h
-    isprime(n) && (h[n] = 1; return h)
+    # check for special cases
+    if n < 0
+        h[-1] = 1
+        if isa(n, BitSigned) && n == typemin(T)
+            h[2] = 8*sizeof(T)-1
+            return h
+        else
+            return factor(checked_neg(n), h)
+        end
+    elseif n == 0
+        h[0] = 1
+        return h
+    elseif n == 1
+        return h
+    elseif isprime(n)
+        h[n] = 1
+        return h
+    end
+
     local p::T
     for p in PRIMES
         if n % p == 0
@@ -254,7 +295,7 @@ function factor{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int}=Dict{T,Int}()
 end
 
 """
-    factor(ContainerType, n) -> ContainerType
+    factor(ContainerType, n::Integer) -> ContainerType
 
 Return the factorization of `n` stored in a `ContainerType`, which must be a
 subtype of `Associative` or `AbstractArray`, a `Set`, or an `IntSet`.
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -207,7 +207,6 @@ euler7(n) = primes(floor(Int,n*log(n*log(n))))[n]
 
 # factor(Vector, n)
 for V in (Vector, Vector{Int}, Vector{Int128})
-    @test_throws ArgumentError factor(V, 0)
     @test factor(V, 1) == Int[]
     @test factor(V, 3) == [3]
     @test factor(V, 4) == [2,2]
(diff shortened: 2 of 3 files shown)


ararslan:
Factoring negative numbers in this way seems okay, though factoring 0 as `0=>1` is kind of odd. Neither 0 nor 1 has a well-defined multiplicity in the sense of factorization. We're still returning an empty container for 1, so why would we return something else for 0? It seems to me that they should be consistent. Personally I'd prefer if 0 and 1 were an error but I'd be interested to hear what others think.

simonbyrne:
+1 This seems reasonable. FWIW, this is what Mathematica does:
```
In[1]:= FactorInteger[-6]                                                                                

Out[1]= {{-1, 1}, {2, 1}, {3, 1}}

In[2]:= FactorInteger[-1]                                                                                

Out[2]= {{-1, 1}}

In[3]:= FactorInteger[0]                                                                                 

Out[3]= {{0, 1}}

In[4]:= FactorInteger[1]                                                                                 

Out[4]= {{1, 1}}
```

src/Primes.jl
```diff
@@ -232,11 +232,35 @@ Dict{Int64,Int64} with 2 entries:
...
+  0 => 1
+```
 """
 function factor{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int}=Dict{T,Int}())
-    0 < n || throw(ArgumentError("number to be factored must be > 0, got $n"))
-    n == 1 && return h
-    isprime(n) && (h[n] = 1; return h)
+    if n < 0
+        h[-1] = 1
+        return factor(-n, h)
```
simonbyrne: Might need to handle `typemin(IntXX)` separately

rfourquet:
Factoring 1 being an error would be very sad, it is standard to consider the empty product to be equal to the multiplicative neutral element of a ring (cf. `prod(Int[]) == 1`), so it's mathematically sound to have `factor(1) == Dict{Int,Int}()`. 

Concerning -1: in a unique factorization domain, the factorization is unique up to multiplication by a unit, i.e. 1 and -1 in the case of integers. Consider the method `factor` extended to factor polynomials (it will necessarily happen, which is why `factor` may be better off defined (i.e. "belonging" to) in Base as a generic function). It would be weird if it was an error to try to factorize `2x`: "error: we can factor only monic polynomials!". So it makes sense, as most math packages which factorize do, to accept any non-null ring elements and output the list of irreducible factors (with a convention, i.e. >0 for integers, monic for polynomials), the multiplicity, and the unit.

For zero it's indeed more controversial! but as there is one obvious way to answer (I would almost dare to say it's canonical), namely `Dict(0=>1)`, I say why not, and let the user decide if she wants to consider that as an error.

The main point is that our invariant is not broken in any case: `prod(factor(Vector, n)) == n`.

src/Primes.jl
rfourquet: Oh indeed! Seems we hit again the problem of `applicable` here, but it should be a much less problematic overhead than in `digits`. But what do you think of inserting `-n < 0 && throw(OverflowError())` in the branch `n<0`. This is similar to `Base.Checked.checked_neg`, but would not force user-defined integers to implement this method (at the cost of maybe being incorrect for those integers). Actually I think I prefer using `checked_neg`.

simonbyrne: Maybe handle it explicitly? i.e.
```julia
if isa(n,Signed) && n == typemin(n)
    h[2] = sizeof(n)*8-1
    return h
end
```

rfourquet:
Updated to use `Base.Checked.checked_neg`.

src/Primes.jl
rfourquet: Ah yes this is a good idea too, but we lose generality as we assume that 2-complement (or something like this) is used... I propose to use this solution but only when we have `Base.BitInteger` types (for lack of a "trait" wich says this is safe to do so).

src/Primes.jl
```diff
@@ -232,11 +232,41 @@ Dict{Int64,Int64} with 2 entries:
...
+```
 """
 function factor{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int}=Dict{T,Int}())
-    0 < n || throw(ArgumentError("number to be factored must be > 0, got $n"))
-    n == 1 && return h
-    isprime(n) && (h[n] = 1; return h)
+    # check for special cases
+    if n < 0
+        h[-1] = 1
+        if isa(n, Base.BitSigned) && n == typemin(T)
```
ararslan: `BitSigned` isn't defined on 0.4, which is what's causing the errors.

ararslan:
Yeah, I agree an error wouldn't be good (even though I said it, haha).

I do think that it would be nice to be consistent with 0 and 1 rather than relying on the product of an empty integer array being 1. Following suit with Mathematica and having `factor(0) == Dict(0=>1)` and `factor(1) == Dict(1=>1)` would make sense to me.

src/Primes.jl
rfourquet: Ah sure, will update.

rfourquet:
It's not really that we rely on the product of an empty array being 1, we rely on the math standard that an empty product is 1. The cases for 0 and 1 are different. In the realm of positive integers, there is an unambiguous definition of unique prime factorization: each number `n>0` can be uniquely factorized as `n = 2^(n_2)*3^(n_3)*5^(n_5)*...` where only a finite number of `n_i` are not null. So the factorization is determined by the list of pairs `(i, n_i)` where `i` is prime and `n_i != 0`. So really for 1 this gives the empty list. Of course Mathematica decided to return `1=>1`, so it could be justified, but I find it odd, please let's not do this! My problem is that it breaks another invariant which I would like to preserve: if `d1, d2` are the factorizations of `n1, n2`, then the "merge" of `d1, d2` is the factorization of `n1*n2` (for an appropriate definition of merge). If `factor(1) == Dict(1=>1)`, then the merge of this with e.g. `factor(5)` would be `factor(1*5) == Dict(1=>1, 5=>1)`...  In other words, our factorization is not really unique anymore (not sure to be convincing here!)

For convenience, this PR proposes to add two new possible keys which are non-primes, to allow to have a "total" function (I think this is the official CS term to describe functions defined on every possible values of a type?), i.e. `-1` and `0`. For `-1` the factorization is still unique and becomes:
every non-null numbers `n` can uniquely be defined as the product `(-1)^t*2^(n_2)*3^(n_3)*5^(n_5)*...` where `t=0` or `t=1`. I agree that this is not much less ugly than the case for `factor(1) = Dict(1=>1)`, but it is necessary to handle negative numbers. For `0` there is no more unicity, so we choose the simplest possible factorization of `0` as the blessed one: `0=0^1`. 
(Sorry for the too long post!)

ararslan:
No need to be sorry, that was a very thorough and convincing explanation. I'm now fully on board with the behavior here. +1, nice work!

ararslan:
Might be worth tagging another release version once this is merged

src/Primes.jl
```diff
@@ -232,11 +237,41 @@ Dict{Int64,Int64} with 2 entries:
...
-    n == 1 && return h
-    isprime(n) && (h[n] = 1; return h)
+    # check for special cases
+    if n < 0
+        h[-1] = 1
+        if isa(n, BitSigned) && n == typemin(T)
+            h[2] = 8*sizeof(T)-1
+            return h
+        else
+            return factor(Base.Checked.checked_neg(n), h)
```
ararslan: `Base.Checked` doesn't exist in 0.4, which is causing the errors

rfourquet: Doh! I need to get used to develop against non-master Julia! Updated, with a hopefully good-enough `checked_neg` for 0.4.

rfourquet:
Glad that you appreciate the behavior here!

test/runtests.jl
```diff
@@ -224,3 +223,13 @@ end
 @test_throws MethodError factor(Int, 10)
 @test_throws MethodError factor(Any, 10)
 @test_throws MethodError factor(Tuple, 10)
+
+# factor non-positive numbers:
+@test factor(0) == Dict(0=>1)
+@test factor(-1) == Dict(-1=>1)
+@test factor(-9) == Dict(-1=>1, 3=>2)
+
+@test factor(typemin(Int)) == Dict(-1=>1, 2=>63)
```
tkelman: remember `Int` is different sizes on 32 vs 64 bit

rfourquet: Ouch thank you!

PR #29, opened 2016-08-30
Lucas-Lehmer(-Riesel) with tests.

vinctux:
Summary:
- Added [Lucas-Lehmer primality test](https://en.wikipedia.org/wiki/Lucas%E2%80%93Lehmer_primality_test) for `M = 2^p - 1` (also known as Mersenne numbers).
- Added [Lucas-Lehmer-Riesel primality test](https://en.wikipedia.org/wiki/Lucas%E2%80%93Lehmer%E2%80%93Riesel_test) for `R = k * 2^n - 1` (not working for arbitrary `k, n` though but [could be solved](https://github.com/JuliaMath/Primes.jl/pull/29#discussion_r77424079)).
- Added tests for LLT and LLRT.


src/Primes.jl
```diff
@@ -384,4 +388,50 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+false
+
+julia> lucaslehmer(13)
+true
+```
+"""
+function lucaslehmer(P::Integer, s::Integer = BigInt(4))
+    if P < 3 return throw(ArgumentError("The condition P ≥ 3 must be met.")) end
+    M = BigInt(2)^P - 1
+    for i in 1:(P - 2) s = (s^2 - 2) % M end
```
simonbyrne: We generally discourage inline for loops.

src/Primes.jl
```diff
@@ -384,4 +388,50 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+julia> riesel(1, 11)  # == lucaslehmer(11)
+false
+
+julia> riesel(3, 607)
+true
+```
+"""
+function riesel(k::Integer, n::Integer)
+    if !(0 < k < BigInt(2)^n) return throw(ArgumentError("The condition 0 < k < 2^n must be met.")) end
+    if k == 1 && n & 1 == 1 return n % 4 == 3 ? lucaslehmer(n, BigInt(3)) : lucaslehmer(n)
```
simonbyrne: And expand these out as well.

src/Primes.jl
```diff
@@ -384,4 +388,56 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+function riesel(k::Integer, n::Integer)
+    0 < k < BigInt(2)^n || return throw(ArgumentError("The condition 0 < k < 2^n must be met."))
+    if k == 1 && n & 1 == 1
+        return n % 4 == 3 ? lucaslehmer(n, BigInt(3)) : lucaslehmer(n)
+    elseif k == 3 && (n % 4) % 3 == 0
+        return lucaslehmer(n, BigInt(5778))
+    elseif (k % 6) % 4 == 1 && (k * 2^n - 1) % 3 > 0
+        pass = 1  # placeholder for later implementation
+    end
+    warn("LLR test is currently not implemented for this form of N.")
```
ararslan: I'd probably make this an error rather than a warning.

src/Primes.jl
```diff
@@ -384,4 +388,56 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+```jldoctest
+julia> lucaslehmer(11)
+false
+
+julia> lucaslehmer(13)
+true
+```
+"""
+function lucaslehmer(P::Integer, s::Integer = BigInt(4))
+    P < 3 && return throw(ArgumentError("The condition P ≥ 3 must be met."))
```
ararslan: Shouldn't use `return` here, just `P < 3 && throw(...)`

src/Primes.jl
```diff
@@ -384,4 +388,56 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+```jldoctest
+julia> riesel(1, 11)  # == lucaslehmer(11)
+false
+
+julia> riesel(3, 607)
+true
+```
+"""
+function riesel(k::Integer, n::Integer)
+    0 < k < BigInt(2)^n || return throw(ArgumentError("The condition 0 < k < 2^n must be met."))
```
ararslan: Same comment regarding `return`

src/Primes.jl
ararslan: I'd also lump the `pass` condition there with the error.

src/Primes.jl
```diff
@@ -384,4 +388,56 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
     end
 end
 
+"""
+    lucaslehmer(P::Integer, [s::Integer = 4]) -> Bool
+
+Lucas-Lehmer primality test for N of form N = 2^P - 1 (aka. Mersenne numbers).
```
ararslan: "Lucas-Lehmer test for primes of the form `2^p - 1`, also known as Mersenne primes."

src/Primes.jl
```diff
@@ -384,4 +388,57 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+    if k == 1 && n & 1 == 1
+        return n % 4 == 3 ? lucaslehmer(n, BigInt(3)) : lucaslehmer(n)
+    elseif k == 3 && (n % 4) % 3 == 0
+        return lucaslehmer(n, BigInt(5778))
+    elseif (k % 6) % 4 == 1 && (k * 2^n - 1) % 3 > 0
+        # TBA
+        error("LLR test is currently not implemented for this form of N.")
+    else
+        error("LLR test is currently not implemented for this form of N.")
+    end
```
ararslan: I'd recommend condensing these two conditions down to a single `else` that throws an exception. For example,
```julia
...
elseif k == 3 && (n % 4) % 3 == 0
    return lucaslehmer(n, BigInt(5778))
else
    # TODO: Implement a case for (k % 6) % 4 == 1 && (k * 2^n - 1) % 3 > 0
    throw(ArgumentError("The LLR test is not currently implemented for numbers of this form."))
end
```
That avoids unnecessary repetition.

ararslan:
I think there's a design decision we have to consider here. For functions like `isprime`, the particular test being used to check for primality isn't part of the API. I think it might be good to be consistent with that and abstract out the implementation to simply `ismersenneprime` rather than `lucaslehmer`. However, I don't know of a common name given to primes of the form `k * 2^n - 1`, which makes `riesel` a little trickier to abstract.

src/Primes.jl
```diff
@@ -384,4 +388,57 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
     end
 end
 
+"""
+    lucaslehmer(P::Integer, [s::Integer = 4]) -> Bool
+
+Lucas-Lehmer primality test for N of form N = 2^P - 1, also known as Mersenne numbers.
```
ararslan: I'd still prefer "Lucas-Lehmer test for primes of the form `2^p - 1`, also known as Mersenne primes." Mersenne *numbers* are natural numbers of the form `n = 2^p - 1`, which can be composite, but this function detects Mersenne *primes*, i.e. Mersenne numbers that are prime.

vinctux: Oh, I intended _primes_ but overlooked it.

src/Primes.jl
```diff
@@ -428,16 +428,17 @@ true
 ```
 """
 function riesel(k::Integer, n::Integer)
-    0 < k < BigInt(2)^n || return throw(ArgumentError("The condition 0 < k < 2^n must be met."))
+    0 < k < BigInt(2)^n || throw(ArgumentError("The condition 0 < k < 2^n must be met."))
     if k == 1 && n & 1 == 1
```
simonbyrne: `n & 1 == 1` is just `isodd(n)` isn't it?

src/Primes.jl
```diff
@@ -428,16 +428,17 @@ true
 ```
 """
 function riesel(k::Integer, n::Integer)
-    0 < k < BigInt(2)^n || return throw(ArgumentError("The condition 0 < k < 2^n must be met."))
+    0 < k < BigInt(2)^n || throw(ArgumentError("The condition 0 < k < 2^n must be met."))
```
simonbyrne: Would be more efficient here to do `0 < k && ndigits(k,2) <= n`

src/Primes.jl
```diff
@@ -428,16 +428,17 @@ true
...
     if k == 1 && n & 1 == 1
         return n % 4 == 3 ? lucaslehmer(n, BigInt(3)) : lucaslehmer(n)
     elseif k == 3 && (n % 4) % 3 == 0
         return lucaslehmer(n, BigInt(5778))
     elseif (k % 6) % 4 == 1 && (k * 2^n - 1) % 3 > 0
-        pass = 1  # placeholder for later implementation
+        # TBA
+        error("LLR test is currently not implemented for this form of N.")
```
simonbyrne: should this be `n` instead of `N`?

src/Primes.jl
```diff
@@ -428,16 +428,17 @@ true
 ```
 """
 function riesel(k::Integer, n::Integer)
-    0 < k < BigInt(2)^n || return throw(ArgumentError("The condition 0 < k < 2^n must be met."))
+    0 < k < BigInt(2)^n || throw(ArgumentError("The condition 0 < k < 2^n must be met."))
     if k == 1 && n & 1 == 1
         return n % 4 == 3 ? lucaslehmer(n, BigInt(3)) : lucaslehmer(n)
     elseif k == 3 && (n % 4) % 3 == 0
         return lucaslehmer(n, BigInt(5778))
     elseif (k % 6) % 4 == 1 && (k * 2^n - 1) % 3 > 0
```
simonbyrne: Do we need to worry about overflow for `(k * 2^n - 1) % 3`?

src/Primes.jl
vinctux: I don't know how costly the init of `BigInt` is but it sounds like a good change, as this is an lonely comparison. I'll change it.

src/Primes.jl
vinctux: Perhaps. But in the latest diff, the `elseif` isn't being evaluated. For the future, a `((k % 3) * powermod(2, n, 3) - 1) % 3 > 0` should do the trick, I guess.

src/Primes.jl
vinctux: Honestly, I don't know. But it seems, at least here, for me on `0.6.0-dev.434`, it seems, that  `isodd(n)` is somehow faster on the long run than `n & 1 == 1`.

```julia
julia> @time isodd(2312312413)
  0.000002 seconds (4 allocations: 160 bytes)
true

julia> @time 2312312413 & 1 == 1
  0.000003 seconds (4 allocations: 160 bytes)
true

julia> @time 2312312413 & 1
  0.00002 seconds (4 allocations: 160 bytes)
1
```
Perhaps because of the low-level code, which is really somewhat different.
```julia
julia> @code_llvm isodd(2312312413)

define i8 @jlsys_isodd_66690(i64) #0 {
top:
  %1 = and i64 %0, 1
  %2 = xor i64 %1, 1
  %3 = trunc i64 %2 to i8
  %4 = xor i8 %3, 1
  ret i8 %4
}

julia> @code_llvm 2312312413 & 1 == 1

define i8 @"jlsys_==_47751"(i64, i64) #0 {
top:
  %2 = icmp eq i64 %0, %1
  %3 = zext i1 %2 to i8
  ret i8 %3
}

julia> @code_llvm 2312312413 & 1

define i64 @"jlsys_&_46390"(i64, i64) #0 {
top:
  %2 = and i64 %1, %0
  ret i64 %2
}
```

simonbyrne: You need to be careful here: firstly, for that low-level of a function the overhead of the function call will outway any realistic benchmarking you can do. Also, to look at the generated code, you should put it in a function, as this lets the compiler do its work:
```
julia> foo(x) = x & 1 == 1
foo (generic function with 1 method)

julia> @code_llvm foo(2312312413)

define i8 @julia_foo_70875(i64) #0 {
top:
  %.tr = trunc i64 %0 to i8
  %1 = and i8 %.tr, 1
  ret i8 %1
}
```
It does look cleaner, and the code_native on my machine appears to be slightly shorter. For this function the difference is going to be negligible compared to the BigInt operations, so I would suggest using `isodd` for ease of comprehension. But you could also open a PR in Base to change the definition (cc me if you do).

vinctux: Thanks for the tips! I'm quite new to Julia, so not a really experienced Julia dev, I guess. I'll change it to `isodd` for the sake of simplicity. Regarding the Base PR - I would love to have the time to dive seriously into helping to make Julia even better but as of this moment, I fear this isn't possible; so I hope somewhere in the near future :)

src/Primes.jl
vinctux: `((k % 3) * powermod(2, n, 3)) % 3 < 2` would further take away an additional operation (even though it's only a substraction by 1).

ararslan:
Turns out that primes of the form *k* 2<sup>*n*</sup> - 1 are called (surprise!) Riesel primes. So I propose the following changes:
* Rename `lucaslehmer` to `ismersenneprime`
* Rename `riesel` to `isrieselprime`

That provides a consistent and (at least IMHO) Julian API.

vinctux:
@ararslan The commit is underway! :)

src/Primes.jl
```diff
@@ -384,4 +388,55 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
     end
 end
 
+"""
+    ismersenneprime(P::Integer, [s::Integer = 4]) -> Bool
+
+Lucas-Lehmer test for primes of the form 2^P - 1, also known as Mersenne primes.
+Returns `true` if 2^P - 1 is prime, and `false` otherwise.
```
ararslan: You know, come to think of it, this function doesn't actually return `true` if the input is a Mersenne prime; it returns `true` if `2^input - 1` is a Mersenne prime. (Right?) Then maybe `ismersenneprime` isn't such an accurate name after all...

vinctux:
Just an off-topic question: how is the README updated? Manually or is there some automation tool in Julia I don't know yet about?

src/Primes.jl
vinctux: IMHO, it's OK how it is now. Passing only the exponent into the function instead of the whole giant prime candidate is just a thing to make it more comfortable for the user. No matter how you spin it - basically seen, the function _does_ determine if the input is a Mersenne prime, it just does it with passing it a "compressed" version of the candidate, because the reproduction is really an easy one.

ararslan: That's a good point.

ararslan:
Since we don't have a documentation system set up yet, we've been updating the documentation in the README manually.

src/Primes.jl
```diff
@@ -6,8 +6,12 @@ if VERSION >= v"0.5.0-dev+4340"
     if isdefined(Base,:isprime)
         import Base: isprime, primes, primesmask, factor
     else
+        # Primal functions (previous Base)
         export isprime, primes, primesmask, factor
     end
+    # Additional functions (non-Base)
```
rfourquet: Oh I understand now where the comments of your other PR comes from! I think you can remove both, we don't need to track in comments which functions were formerly in base.

src/Primes.jl
```diff
@@ -384,4 +388,55 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
     end
 end
 
+"""
+    ismersenneprime(P::Integer, [s::Integer = 4]) -> Bool
+
```
rfourquet: The parameter `s` is not documented... For which values of `s` the test is proven to be valid?
Also, I would constrain `P::Int` (possibly `P::Integer as Int` in the future if this syntax for automatic conversion becomes available), and use lower case `p` instead, as this file already often uses `p` for primes.

src/Primes.jl
rfourquet: I'm not totally sold on this. As you said, if the user wants an "easier" fonction, it's nothing for her to add a wrapper. The advantage of accepting directly `M = 2^p-1` as a parameter is that it allows to work with possibly faster-than-BigInt integers. Also, even if you are right that passing `p` is a way to encode the Mersenne number, it's still a bit confusing that `isprime` and `ismersenneprime` would use different encondings. 
I would suggest to have `ismersnneprime` accept `M` directly, and why not include `lucaslehmer(p)` which accepts the exponent.

src/Primes.jl
```diff
@@ -384,4 +388,55 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+julia> ismersenneprime(13)
+true
+```
+"""
+function ismersenneprime(P::Integer, s::Integer = BigInt(4))
+    P < 3 && throw(ArgumentError("The condition P ≥ 3 must be met."))
+    M = BigInt(2)^P - 1
+    for i in 1:(P - 2)
+        s = (s^2 - 2) % M
```
rfourquet: On wikipedia, an optimization is mentioned to compute this modulo operation, do you have an idea how much it would improve performances here? maybe add a TODO refering to this possible optimization.

src/Primes.jl
```diff
@@ -384,4 +388,55 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+```jldoctest
+julia> ismersenneprime(11)
+false
+
+julia> ismersenneprime(13)
+true
+```
+"""
+function ismersenneprime(P::Integer, s::Integer = BigInt(4))
+    P < 3 && throw(ArgumentError("The condition P ≥ 3 must be met."))
```
rfourquet: It user-unfriendly, throw only if `p<0` (or `p<1` if you wish), and for other small powers just hardcode the result.

src/Primes.jl
```diff
@@ -384,4 +388,55 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+    for i in 1:(P - 2)
+        s = (s^2 - 2) % M
+    end
+    return s == 0
+end
+
+"""
+    isrieselprime(k::Integer, n::Integer) -> Bool
+
+Lucas-Lehmer-Riesel primality test for N of form N = k * 2^n - 1,
```
rfourquet: "of form" -> "of the form"

src/Primes.jl
```diff
@@ -384,4 +388,55 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+```jldoctest
+julia> isrieselprime(1, 11)  # == ismersenneprime(11)
+false
+
+julia> isrieselprime(3, 607)
+true
+```
+"""
+function isrieselprime(k::Integer, n::Integer)
+    0 < k && ndigits(k,2) ≤ n || throw(ArgumentError("The condition 0 < k < 2^n must be met."))
```
rfourquet: `ndigits(k, 2)`, i.e. add a space to be consistent with your other PR ;-)

src/Primes.jl
```diff
@@ -384,4 +388,55 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+    if k == 1 && isodd(n)
+        return n % 4 == 3 ? ismersenneprime(n, BigInt(3)) : ismersenneprime(n)
+    elseif k == 3 && (n % 4) % 3 == 0
+        return ismersenneprime(n, BigInt(5778))
+    else
+        # TODO: Implement a case for (k % 6) % 4 == 1 && ((k % 3) * powermod(2, n, 3)) % 3 < 2
```
rfourquet: I didn't take the time yet to study this test, but how difficult is it to implement this TODO? Why not wait to implement it fully before including this function here with another PR? (i.e. it can be frustrating for users if we advertise to implement this test, but it throws on many inputs).

src/Primes.jl
simonbyrne: Another possible solution is to define a `Mersenne` type which is stored as the exponent:
```julia
immutable Mersenne <: Integer
   n::Int
end
```
and then dispatch `isprime(x::Mersenne)`.

vinctux: @rfourquet I agree, that accepting the whole `M = 2^p-1` would certainly _be_ better for faster-than-BigInt integers but that fact applies to the computation cycle only, as you must first run a check if the given param is a valid `M`, which would IIRC require something like `O(log M)` logical shifts to find out. It would cause the function to be even slower on bigger structures, like `BigInt`, as it would require even longer to get the exponent from the candidate.

So, I guess that the only _real_ advantage of accepting the whole candidate instead of the exponent version is merely for the user friendliness.

src/Primes.jl
vinctux: Yeah, those are really obsolete and I'll remove them in the next commit.

src/Primes.jl
vinctux: I haven't implemented the case because it does require exponentiation of an irrational, which can certainly point to problems. The test can also be rewritten as a solution to a set equalities with Jacobi's symbol, which would nearly certain require a dependency on a package as `Combinatorics.jl` or something similar, so for now, I didn't want to introduce a new dependency in this a PR. We can always add a new dependency and generalise the test for _all_ `(k, n)` but I hadn't done this as of yet.

vinctux:
Ew, just committed when the new codestyle was merged. Will have to do another commit :P

vinctux:
The `49e7c23` merge commit will fail on CIs, because I had done some mess with git there. The other one should pass without problems, runs for me, at least :)

vinctux:
Now I see... All new types aren't imported in `runtests.jl`. That's why it fails on pre 0.5 versions.

src/Primes.jl
```diff
@@ -9,6 +9,11 @@ if VERSION >= v"0.5.0-dev+4340"
         export isprime, primes, primesmask, factor
     end
 
+    export MersenneNumber, RieselNumber
+    export Mersenne, Riesel
+    export BigMersenne, BigRiesel
+    export mersenne, riesel, bigmersenne, bigriesel
```
ararslan: The `export`s should go outside of the condition on the version, otherwise it won't work on 0.4

src/Primes.jl
```diff
@@ -392,4 +397,167 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+immutable BigMersenne <: MersenneNumber
+    p::BigInt
+
+    """
+        BigMersenne(p::Integer) -> BigMersenne
+
+    Creates a `BigMersenne` object, given the argument `p`, as in `M = 2^p - 1`.
+    Throws an error if `p` is not prime.
+    """
+    BigMersenne(p::Integer) = isprime(p) ? new(BigInt(p)) : throw(ArgumentError("The argument given is not a prime."))
```
vinctux: I can't spot the erroneous part of code. Maybe my eyes are just too tired...

```julia
julia> BigMersenne(11) == bigmersenne(2^11-1)
false

julia> Mersenne(11) == mersenne(2^11-1)
true
```

I updated the tests, so the coverage won't drop with the merge but `BigMersenne` and `BigRiesel` are causing problems, apparently.

src/Primes.jl
```diff
@@ -392,4 +397,167 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+Throws an error if given `Q` is not of the form 2^n - 1.
+"""
+function riesel(k::Integer, Q::Integer)
+    b = bin(Q)
+    '0' in b && throw(ArgumentError("The arguments given do not result in a valid Riesel number.")) || (n = length(b))
+    return Riesel(k, n)
+end
+
+### Big types ###
+immutable BigMersenne <: MersenneNumber
```
rfourquet: Typically, you would use only one type with a parameter encoding the type of `p`.

rfourquet:
Thanks for your hard work. Personally I'm not conviced I like to introduce a MersenneNumber type. I would rather have e.g. `ismersenneprime(M)` which calls `ismersenneprime_unchecked(M)` if we really want to check that `M` is a valid Mersenne number. But in my optinion, it would be enough to have `ismersenneprime` unchecked by default, as it's such a specialized function that users of it will use it with care. The doc could be: "if the argument `M` is Mersenne number, then return `true` iff `M` is a prime number."

vinctux:
As I understand the argument of yours, it should be kept in mind, that I had written `Mersenne` and `Riesel` only to satisfy the requirement of giving the user the option to choose between passing the whole number *and* the exponent alone. If I understand correctly, you're proposing a secondary function `ismersenneprime_unchecked(M)` for _not_ checking if the given argument (which I consider to be the whole number, ergo `M = 2^p + X`) is really a valid Mersenne number by definition (`X = -1`). IMHO, this proposal is somehow unrelated to the original topic on how to allow the user to pass the _same_ argument in two distinct ways, without consuming too many resources. @simonbyrne [proposed the `Mersenne` immutable](https://github.com/JuliaMath/Primes.jl/pull/29#discussion_r77167311), which I think is a fairly good solution to this "problem", so I implemented it and expanded the idea to `Riesel`. With keeping a low profile on resources in mind, I additionally added `BigMersenne` and `BigRiesel`, which store and process their (coefficients and) exponents as `BigInt`s, whereas `Mersenne` uses `Int16` (which should be sufficient for Mersenne numbers up until `2^65536 - 1`) and `Riesel` uses `Int64` for `k`, `Int16` for `n`. So, basically the "Big" types were my first idea, and `Mersenne` and `Riesel` exist, as they do now, as a low-profile variant if the user is dealing with relatively small numbers.

I do not like the idea of creating a new type for every number with certain properties, which can vastly speed up computation in some circumstances, either. To solve _this_ issue, a change in Base would be required, adding a new generic type of numbers, with additional fields for their properties. It would require some work to implement it and even more beforehand, to correctly design the structure for such a system -- it would certainly be a difficult task to tackle with. But, as I see it, this is currently out of scope for this very PR.

If you have any further suggestions or opinions, feel free. I'm more than fine to change the code, as needed.

rfourquet:
As you suggested, I fear that if we go this route, we will end up with a cluttered repo with many different number types encoding special classes. And those types aren't even integer types (and can not interact with standard integer types directly). I think that this adds too much maintenance overhead for something which is really useful to probably only a small handful of people. For the majority of those who need to use those 2 tests (which as of now are not that many among julia users, I guess), unchecked versions of `ismersenneprime` and `isrieselprime` are plenty sufficient, provided clear documentation.

When we have more experience and other candidates, we *may* think of adding specialized number types (In a private repo, I introduced such numbers (in a totally different topic), and am now thinking of removing that, as the benefit is too low compared to the maintenance overhead, in my case).

So my vote is to only have those simple two functions, taking the argument `M` directly (unchecked, if this is measurably faster). I believe those could eventually be merged to `isprime` via keyword arguments (e.g. `isprime(M, ismersenne=true)` (or `test=lucaslehmer`)).

ararslan:
I agree with @rfourquet, though I'm not convinced about merging them into `isprime`. But that's a separate issue.

vinctux:
@rfourquet For the current state of play, I agree that, as I said earlier, it would create maintenance overhead, while providing functionality for a minority. Although your proposed reversal is IMO a short-hand solution to a more massive issue, for now, the keyword arguments should be sufficient. The long-term approach would be a complete reconstruction of `Base`, from ground up, dynamically allocating bits for specific properties (Fermat number, Mersenne number, perfect square, etc.), which could be evaluated and assigned on per-property basis. But that's, as said, a far too radical choice in the scope of this PR.

I'll revert the `Mersenne` and Co. shortly. Any other changes needed to get this merged?

src/Primes.jl
```diff
@@ -2,16 +2,21 @@
...
         import Base: isprime, primes, primesmask, factor
     else
         export isprime, primes, primesmask, factor
     end
 
+    for s in instances(PrimeTest) @eval export $(Symbol(s)) end
+
     using Base: BitSigned
     using Base.Checked.checked_neg
 else
```
ararslan: I don't think we should do this (in general, or at least not now). IMO we should stick to `ismersenneprime` and `isrieselprime` and maybe revisit something like this in the future.

src/Primes.jl
```diff
@@ -137,16 +145,25 @@ primes(n::Int) = primes(1, n)
 const PRIMES = primes(2^16)
 
 """
-    isprime(x::Integer) -> Bool
+    isprime(n::Integer; [k::Integer = 0, T::PrimeTest = gentest]) -> Bool
```
ararslan: I don't think we should do this (in general, or at least not now). IMO we should stick to `ismersenneprime` and `isrieselprime` and maybe revisit something like this in the future.

vinctux:
This is getting serious `git` mess here... At merge all these commits should be squeezed into one, dropping the description (á la `fixup`), as they're fixes out of discussions from this PR.

Under Julia 0.4 the export of `PrimeTest` enum fails and I don't why exactly. Also, the coverage drop is mysterious to me, as no untested code had been added.

ararslan:
Looks like the consensus is to keep it as `ismersenneprime` and `isrieselprime` for now without modifying `isprime`, so the enum failure should be a non-issue once that's addressed.

Don't worry too much about Git messiness; I figured we'd squash on merge.

src/Primes.jl
rfourquet: Thank you for prototyping this, but as @ararslan said I think it's a bit premature. Let's stay conservative with this PR, by focusing on the implementation of `ismersenneprime` and `isrieselprime`. We *may* have a future PR to introduce those keywords arguments, but only when we have a little more algorithms available, and probably some use cases (e.g. requests from users).

vinctux:
Hi again! It's been some time, I was busy and nearly forgot about this PR - here's now the long awaited commit, hopefully getting it through the checks and merged. :smile: The whole PR is a mess of 10 commits - so to anyone who's about to look for it before merge - I'm sorry for it. The last commit required a manual reverse of the prototype, as the prototyping was committed together with other critical changes.
Tested on `0.6.0-dev.1276`, couldn't run the tests for 0.4 or 0.5, though; my new install has apparently some problems with those versions... so there certainly will be (hopefully not too much) backward incompatible things noticed by Travis.

LGTFM! :facepunch:

src/Primes.jl
```diff
@@ -392,4 +395,64 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+        return ll_primecheck(Q, 5778)
+    else
+        # TODO: Implement a case for (k % 6) % 4 == 1 && ((k % 3) * powermod(2, n, 3)) % 3 < 2
+        error("The LLR test is not currently implemented for numbers of this form.")
+    end
+end
+
+# LL backend -- not for export
+function ll_primecheck(X::Integer, s::Integer = 4)
+    S, N = BigInt(s), BigInt(length(bin(X)))
```
ararslan: `ndigits(X, 2)` should be more efficient than `length(bin(X))`

src/Primes.jl
```diff
@@ -392,4 +395,64 @@ function pollardfactors!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
+```jldoctest
+julia> isrieselprime(1, 2^11 - 1)  # == ismersenneprime(2^11 - 1)
+false
+
+julia> isrieselprime(3, 2^607 - 1)
+true
+```
+"""
+function isrieselprime(k::Integer, Q::Integer)
+    n = length(bin(Q))
```
ararslan: here too

src/Primes.jl
```diff
@@ -20,6 +20,9 @@ else
     end
 end
 
+export ismersenneprime, isrieselprime
+
+# Prime tests enum and export
```
ararslan: Stray or otherwise leftover comment?

src/Primes.jl
```diff
@@ -137,9 +138,9 @@ primes(n::Int) = primes(1, n)
 const PRIMES = primes(2^16)
 
 """
-    isprime(x::Integer) -> Bool
+    isprime(n::Integer) -> Bool
 
-Returns `true` if `x` is prime, and `false` otherwise.
+Returns `true` if `n` is prime, and `false` otherwise.
```
ararslan: Why change this?

vinctux: _x_ in mathematics usually denotes an element from ℝ - a real number, where _n_ usually denotes an element from ℕ - a natural number. While it holds that ℙ ⊂ ℕ ⊂ ℝ, nonetheless ℙ is far more closely related to ℕ as to ℝ. Also the function `isprime` accepts `Integer`, which aims to represent ℕ, not ℝ. It's a purely cosmetic change, though.

TL;DR: The user will never pass a real number _x_ into `isprime` but a natural number _n_.

ararslan: Since the change is unrelated to the PR, please remove it; that change can be revisited later.

vinctux: Shall I reverse it anyway?

ararslan: Nah, whatevs. Not a huge deal.

vinctux:
Any update on this PR? It's been 4 months since the approval - if anything needs to be changed before a merge, I'm here, all ear ;)

ararslan:
Apparently no one else is going to comment, so I'll just go ahead and merge. Thanks for the reminder, and of course for the contribution.

PR #30, opened 2016-08-30
Improved code readability.

vinctux:
Summary:

- Inserted spaces for greater readability of code.
- Compressed statements (like `r = 2 * r` into `r *= 2`).
- Added `return` keywords.
- Improved text of errors.
- Regrouped `bases` at 12 bases at a line.


src/Primes.jl
```diff
@@ -30,26 +32,27 @@ const wheel_primes  = [7, 11, 13, 17, 19, 23, 29, 31]
...
 @inline function wheel_prime(n)
-    (d, r) = ((n - 1) >>> 3, (n - 1) & 7)
-    30d + wheel_primes[r+1]
+    d, r = (n - 1) >>> 3, (n - 1) & 7
+    return 30d + wheel_primes[r + 1]
 end
 
 function _primesmask(limit::Int)
-    limit < 7 && throw(ArgumentError("limit must be at least 7, got $limit"))
+    limit < 7 && throw(ArgumentError("The condition limit ≥ 7 must be met."))
```
ararslan: I actually prefer the existing error message.

src/Primes.jl
```diff
@@ -100,27 +105,27 @@ function primesmask(lo::Int, hi::Int)
...
 """
     primes([lo,] hi)
 
 Returns a collection of the prime numbers (from `lo`, if specified) up to `hi`.
 """
 function primes(lo::Int, hi::Int)
-    lo ≤ hi || throw(ArgumentError("the condition lo ≤ hi must be met"))
-    list = Int[]
+    lo ≤ hi || throw(ArgumentError("The condition lo ≤ hi must be met."))
+    list = Array{Int, 1}()
```
ararslan: `Int[]` is a little more readable, IMO

src/Primes.jl
```diff
@@ -187,23 +192,22 @@ isprime(x::BigInt, reps=25) = ccall((:__gmpz_probab_prime_p,:libgmp), Cint, (Ptr
...
+                      1661,   524,  3577,   236,  2360,  6146,  2850, 55637,  1753,  4178,  8466,   222,  2579,  2743,  2031,  2226,
+                      2276,   374,  2132,   813, 23788,  1610,  4422,  5159,  1725,  3597,  3366, 14336,   579,   165,  1375, 10018,
+                     12616,  9816,  1371,   536,  1867, 10864,   857,  2206,  5788,   434,  8085, 17618,   727,  3639,  1595,  4944,
+                      2129,  2029,  8195,  8344,  6232,  9183,  8126,  1870,  3296,  7455,  8947, 25017,   541, 19115,   368,   566,
+                      5674,   411,   522,  1027,  8215,  2050,  6544, 10049,   614,   774,  2333,  3007, 35201,  4706,  1152,  1785,
+                      1028,  1540,  3743,   493,  4474,  2521, 26845,  8354,   864, 18915,  5465,  2447,    42,  4511,  1660,   166,
+                      1249,  6259,  2553,   304,   272,  7286,    73,  6554,   899,  2816,  5197, 13330,  7054,  2818,  3199,   811,
+                       922,   350,  7514,  4452,  3449,  2663,  4708,   418,  1621,  1171,  3471,    88, 11345,   412,  1559,   194]
```
ararslan: In the interest of respecting line lengths, what would you think of
```julia
const bases = UInt16[
    11591,  2018, ...
     2300,  2021, ...
    ...
]
```
?

src/Primes.jl
```diff
@@ -234,35 +238,32 @@ function factor!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
         else
             return factor!(checked_neg(n), h)
         end
-    elseif n == 0
-        h[0] = 1
+    elseif n == 0 || isprime(n)
+        h[n] = 1
         return h
     elseif n == 1
         return h
```
ararslan: I wonder if the `n == 1` condition should be moved atop `n == 0 || isprime(n)` so that `isprime` doesn't have to deal with 1. Dunno if it'll actually make a difference but at least it avoids a function call in that case.

src/Primes.jl
```diff
@@ -337,48 +338,48 @@ Set([2,5])
...
             end
-            r = 2 * r
+            r *= 2
         end
         G == n && (G = 1)
         while G == 1
-            ys = (ys*ys)%n
-            ys = (ys+c)%n
-            G = gcd(x>ys?x-ys:ys-x,n)
+            ys = ys^2 % n
```
ararslan: Does `^` have higher precedence than `%`? Also, is there any performance penalty for exponentiation versus multiplication? I think this is fine how you've done it but I just wanted to check.

vinctux: [Yes](http://docs.julialang.org/en/latest/manual/mathematical-operations/#operator-precedence). [No](https://github.com/JuliaLang/julia/blob/master/base/intfuncs.jl#L143). :-)

src/Primes.jl
vinctux: Perhaps `Int[]` is really more readable but I thought of the change as a more long-term 'fix' if `Int[]` (which is very generic writing, without any unique meaning given) gets to mean something, which isn't exactly a one-dimensional array of `Int`, which is what the function expects. But in the current state, the change can be regarded obsolete, though.

src/Primes.jl
vinctux: This should further minimise unnecessary code execution. I'll change it.

src/Primes.jl
vinctux: I'll change it. It's 116 chars/line, though.

simonbyrne: FWIW, Julia itself has a 92 character limit, which I've found to be quite a good limit.

src/Primes.jl
ararslan: That's a good point. Might as well go with `Array` then. Could be even more explicit as `Vector{Int}()`, but I'm fine with `Array{Int,1}()` if you prefer it.

src/Primes.jl
ararslan: If we restrict the number of columns in this "matrix" to 12, we can get the line length down to 88. However, doing so could be an incredibly tedious task, unless you're a Vim wizard or something. :smile:

src/Primes.jl
vinctux: I haven't looked into it yet - are there any real differences between `Array{Int, 1}()` and `Vector{Int}()` or is `Vector{T}()` just a shorthand for `Array{T, 1}`?

ararslan: `Vector{T}` is just a convenient `typealias` for `Array{T,1}`, so they're entirely equivalent.

My only beef the code here as you have it is the space after the comma. IMO `Array{T,1}` looks better than `Array{T, 1}`, but it's just personal preference, not a big deal at all.

vinctux: Should we go for `Vector{T}` then?

ararslan: I like `Vector{T}` but I'll leave the final call on that up to you. :wink:

vinctux: Thanks. I guess I'll leave it as it is now, not making microscopic commits that load up Travis for an hour. If there will be any need for it, I'll add it just before the merge, combining all ideas that will pop up until them.

ararslan:
This looks good to me, apart from the line length issue with `bases`.

vinctux:
So, groups of 8 rather than 16?

ararslan:
Could even do groups of 12 for a total line length of ~88 (close to base Julia's limit of 92)

vinctux:
OK, done. Any other changes before I commit and heavy load Travis?

ararslan:
Looks good to me! I have no other comments so I think this is good to go, pending feedback from others.

Also, it's great that you're mindful of the Travis load but you don't need to worry *too* much here; we don't share a queue with JuliaLang and each build of this package only takes between 2 and 7 minutes. So we can afford some extra commits as needed. 😉

src/Primes.jl
```diff
@@ -6,16 +6,18 @@ if VERSION >= v"0.5.0-dev+4340"
     if isdefined(Base,:isprime)
         import Base: isprime, primes, primesmask, factor
     else
+        # Primal functions (previous Base)
```
rfourquet: Actually this comment appears confusing to me.

src/Primes.jl
```diff
@@ -30,26 +32,27 @@ const wheel_primes  = [7, 11, 13, 17, 19, 23, 29, 31]
...
+    limit < 7 && throw(ArgumentError("The condition limit ≥ 7 must be met."))
     n = wheel_index(limit)
     m = wheel_prime(n)
     sieve = ones(Bool, n)
     @inbounds for i = 1:wheel_index(isqrt(limit))
-        if sieve[i]; p = wheel_prime(i)
-            q = p * p
+        if sieve[i]
+            p = wheel_prime(i)
+            q = p^2
```
rfourquet: Isn't there a performance argument for keeping `p*p` instead of `p^2` ?

src/Primes.jl
simonbyrne: > Perhaps Int[] is really more readable but I thought of the change as a more long-term 'fix' if Int[] (which is very generic writing, without any unique meaning given

I'm not sure what you mean here. Julia is very clear that `T[]` is an empty `Vector{T}`, and this is used quite a bit in Base.

src/Primes.jl
simonbyrne: No, `p^2` is inlined to `p*p`, so it's merely a stylistic thing.

src/Primes.jl
rfourquet: I too don't understand this change.

src/Primes.jl
rfourquet: Ah sorry for my redundant comment above, I had not reached here yet... But the link to `power_by_squaring` does not convince me 100%, as this code (for `x^2`) contains some branches before returning `x*x`. Can you run some benchmark to be sure?

test/runtests.jl
```diff
@@ -202,17 +200,17 @@ end
...
+euler7(n) = primes(floor(Int, n * log(n * log(n))))[n]
 @test euler7(10001) == 104743
 
 # project euler 10: 142913828922
 @test sum(map(Int64,primes(2000000))) == 142913828922
 
 # factor(Vector, n)
 for V in (Vector, Vector{Int}, Vector{Int128})
-    @test factor(V, 1) == Int[]
+    @test factor(V, 1) == Array{Int, 1}()
```
rfourquet: Same as above, I would keep `Int[]`.

src/Primes.jl
simonbyrne: On 0.5-rc3:
```julia
ulia> foo(x) = x^2
foo (generic function with 1 method)

julia> @code_llvm foo(1)

define i64 @julia_foo_70519(i64) #0 {
top:
  %1 = mul i64 %0, %0
  ret i64 %1
}
```

src/Primes.jl
```diff
@@ -30,26 +32,27 @@ const wheel_primes  = [7, 11, 13, 17, 19, 23, 29, 31]
 const wheel_indices = [0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 3, 3, 3, 3, 4, 4, 5, 5, 5, 5, 6, 6, 6, 6, 6, 6, 7, 7]
 
 @inline function wheel_index(n)
-    (d, r) = divrem(n - 1, 30)
-    8d + wheel_indices[r+2]
+    d, r = divrem(n - 1, 30)
+    return 8d + wheel_indices[r + 2]
 end
```
rfourquet: I like adding space for readability, but I don't feel it's always necessary for simple algebraic expressions like here. I personally use to put space to express an operation which binds less tightly, e.g. `3*r + 2`. I think when there is not an obvious better style, we tend to keep it as is (as the primary author decided to write). But of course I'm fine with this harmless change.

src/Primes.jl
rfourquet: Ok great!

src/Primes.jl
vinctux: Should it be removed?

src/Primes.jl
vinctux: Should it be changed back, then?

ararslan: Yeah, probably best to leave it as it was.

src/Primes.jl
ararslan: Probably. I also don't find it particularly descriptive. It's clear that it used to be in Base by virtue of being in the `else` block when checking whether it's defined in Base. :smile:

src/Primes.jl
vinctux: OK, no problem, I'll revert it.

src/Primes.jl
vinctux: I'll remove it with the next commit. Anyway, scrolling through my old `git log`, it appears that the comment wasn't even intended to be included in this merge - so, basically an error from my side. It's confusing indeed.

PR #31, opened 2016-10-03
WIP: Add docs with Documenter.jl

ararslan:
This is my first time using Documenter so let me know if there's anything that looks incorrect, though everything builds just fine locally.

To do:
- [ ] The functions show up as `Base.isprime`, etc., despite being explicitly prefixed in docs/src/api.md
- [x] Set up automatic updating of docs when changes are pushed
- [x] Figure out where to host the docs


docs/src/index.md
```diff
@@ -0,0 +1,26 @@
...
+To install it, run
+```julia
+Pkg.add("Primes")
+```
+from the Julia REPL.
+
+## Note
+
+Prior to Julia 0.5, these (or similar) functions were available in Julia's Base module.
+Because of this, the symbols from this package are not exported on Julia 0.4 to avoid
```
waldyrious: Minor suggestion: changing to active voice ("the symbols from this package are not exported" --> "this package does not export its symbols") would make this sentence clearer, IMO.

ararslan: I actually kind of prefer it as is, to be honest, though I appreciate your input.

waldyrious: Sure, it's not a big deal. I understand it may sound less natural, but for a moment I got confused about Julia not exporting these symbols, rather than this package.

rfourquet:
Thanks for having tackled this, this is great! It was really annoying before to have to duplicate the docstrings in the readme (and with an error-prone syntax, at least for me).

ararslan:
No problem! That reminds me, I should update this for the latest Documenter practices.

PR #33, opened 2016-12-21
Replace deprecated xor operator with unicode U+22BB xor character

Zyndric:
Fixes a warning that Julia dev issues for use of the recently deprecated dollar operator.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.4
+Compat 0.9.4
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -217,9 +217,9 @@ const bases = UInt16[
 ]
 
 function _witnesses(n::UInt64)
-    i = ((n >> 16) $ n) * 0x45d9f3b
-    i = ((i >> 16) $ i) * 0x45d9f3b
-    i = ((i >> 16) $ i) & 255 + 1
+    i = ((n >> 16) ⊻ n) * 0x45d9f3b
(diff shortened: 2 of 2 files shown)


ararslan:
Hi @Zyndric! Thanks for the patch. This change will only work on Julia 0.6, since the v underbar character doesn't have a meaning in Julia 0.4 and 0.5. To fix this, we'll need to do two things:
* Add a dependency on Compat.jl, setting the minimum required version to the version that introduced the `xor` change. To do this, add `Compat 0.9.4` to REQUIRE underneat `julia 0.4`.
* Add `using Compat` beneath the `module Primes` statement in src/Primes.jl

That should be it! If you could make those changes as part of this PR, that would be fantastic.

ararslan:
The `xor` function is similarly backwards incompatible; `$` was the only XOR operator until the PR which introduced both v underbar and `xor`.

Zyndric:
Haha, thanks for your comment. I just read it after making the other commit. Anyway, which of the two operators would you rather see?

ararslan:
Personally I don't have much of a preference. Since you've already added the commit that uses `xor`, we can just go with that one. 🙂

Zyndric:
BTW, where did you see that Compat 0.9.4 introduced the xor operator change, and not another version?

ararslan:
I looked at the Compat PR which introduced it and compared the date to the date of the next subsequent Compat release.

ararslan:
Thanks!

PR #37, opened 2017-04-18
add ismersenneprime to docs

rfourquet:
I don't know much about Documenter, not sure this change is enough...


ararslan:
Yep, that does it. Thanks!

PR #38, opened 2017-04-18
factor returns a sorted factorization object

rfourquet:
This is inspired from the sagemath system (cf. http://doc.sagemath.org/html/en/reference/structure/sage/structure/factorization.html). This factorization object is an associative containter which prints nicely and is sorted:
```
julia> factor(100)
2^2 ⋅ 5^2
```
Note that the output can be pasted back in the REPL to give the product.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -8,6 +8,8 @@ if isdefined(Base, :Iterators)
     import Base.Iterators: repeated
 end
 
+include("factorization.jl")
+
 if VERSION >= v"0.5.0-dev+4340"
     if isdefined(Base,:isprime)
         import Base: isprime, primes, primesmask, factor
@@ -282,35 +284,35 @@ end
 
 
 """
-    factor(n::Integer) -> Dict
+    factor(n::Integer) -> Primes.Factorization
 
-Compute the prime factorization of an integer `n`. Returns a dictionary. The
-keys of the dictionary correspond to the factors, and hence are of the same type as `n`.
-The value associated with each key indicates the number of times the factor appears in the
-factorization.
+Compute the prime factorization of an integer `n`. Return a sorted
+dictionary (with specialized printing) whose keys correspond to the
+factors, and hence are of the same type as `n`. The value associated
+with each key indicates the multiplicity (i.e. the number of times the
+factor appears in the factorization).
 
 ```julia
-julia> factor(100) # == 2^2 * 5^2
-Dict{Int64,Int64} with 2 entries:
-  2 => 2
-  5 => 2
+julia> factor(100)
+2^2 ⋅ 5^2
 ```
 
 For convenience, a negative number `n` is factored as `-1*(-n)` (i.e. `-1` is considered
 to be a factor), and `0` is factored as `0^1`:
 
 ```julia
-julia> factor(-9) # == -1 * 3^2
-Dict{Int64,Int64} with 2 entries:
-  -1 => 1
-   3 => 2
+julia> factor(-9)
+-1 ⋅ 3^2
 
 julia> factor(0)
-Dict{Int64,Int64} with 1 entries:
-  0 => 1
+0
+
+julia> collect(factor(0))
+1-element Array{Pair{Int64,Int64},1}:
+ 0=>1
 ```
 """
-factor{T<:Integer}(n::T) = factor!(n, Dict{T,Int}())
+factor{T<:Integer}(n::T) = factor!(n, Factorization{T}())
 
 
 """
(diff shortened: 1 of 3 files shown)


src/factorization.jl
```diff
@@ -0,0 +1,46 @@
+# implementation of a sorted dict (not optimized for speed) for storing
+# the factorization of an integer
+
+immutable Factorization{T<:Integer} <: Associative{T, Int}
+    pe::Vector{Pair{T, Int}} # Prime-Exponent
+
+    # Factorization{T}() where {T} = new(Vector{Pair{T, Int}}())
+    (::Type{Factorization{T}}){T<:Integer}() = new{T}(Vector{Pair{T, Int}}())
```
ararslan: There isn't really a compelling case for supporting 0.4 at this point, but the fix to get it working there is simple enough: this and the definition below just need a `@compat`. (Plus the package needs a dependency on Compat, if we don't already have one.)

rfourquet: Ok thank you for the hint, updated. Except that printing doesn't work as intended (the `Factorization` object prints as a normal dictionary), though it doesn't seem worth it to fix that. As 0.4 had Primes in Base, it's indeed not compelling to support 0.4 here. Let's stop its support when 0.6 is out!

ararslan: > Except that printing doesn't work as intended (the Factorization object prints as a normal dictionary)

You mean on 0.4? That's weird... I haven't seen anything in here that's obviously not 0.4-friendly now that you've added the `@compat`s. Oh well, I agree that it's not worth fixing.

ararslan:
It would be great to have more tests for the various things that are overloaded for `Factorization`.

rfourquet:
Good idea, I added basic tests. 
@simonbyrne, @pabloferz what is your opinion on this change?

pabloferz:
:+1:  Love this idea. Also, seems to be ready on my end.

src/factorization.jl
```diff
@@ -0,0 +1,46 @@
+# implementation of a sorted dict (not optimized for speed) for storing
+# the factorization of an integer
+
+immutable Factorization{T<:Integer} <: Associative{T, Int}
```
simonbyrne: This does clash with the name `Factorization` in Base. Not sure how big of an issue that could be.

src/Primes.jl
```diff
@@ -282,35 +284,35 @@ end
...
 """
-    factor(n::Integer) -> Dict
+    factor(n::Integer) -> Primes.Factorization
 
-Compute the prime factorization of an integer `n`. Returns a dictionary. The
-keys of the dictionary correspond to the factors, and hence are of the same type as `n`.
-The value associated with each key indicates the number of times the factor appears in the
-factorization.
+Compute the prime factorization of an integer `n`. Return a sorted
+dictionary (with specialized printing) whose keys correspond to the
```
simonbyrne: Should be "Return a `Factorization` object"

src/factorization.jl
ararslan: Might be kind of lengthy, but `PrimeFactorization` could be a descriptive name to use here to avoid clashing with Base types.

pabloferz: If this one is not exported I don't think it would be much of an issue.

simonbyrne: Actually thinking more about it, it's probably fine: this is the reason we have namespaces after all.

src/Primes.jl
ararslan: Pushed a commit to the branch that addresses this.

rfourquet: I hesitated about that, the signature already states that the returned object of type `Factorization`, and still this a (kind of) dictionary.

src/factorization.jl
rfourquet: To be honest, I even thought about whether those two factorization objects could be unified. Or maybe `Factorization` should be an abstract type? anyway, indeed I prefer to keep here the namespaced name as is.

src/factorization.jl
rfourquet: Yes 0.4. I think it's the way `show` woks which has been changed, I don't remember the correct invocations for 0.4.

src/Primes.jl
```diff
@@ -286,8 +286,8 @@ end
 """
     factor(n::Integer) -> Primes.Factorization
 
-Compute the prime factorization of an integer `n`. Return a sorted
-dictionary (with specialized printing) whose keys correspond to the
+Compute the prime factorization of an integer `n`. Returns a `Factorization`
```
rfourquet: I had corrected `Returns` to `Return` in my first commit, for consistency (cf. "Compute" in previous sentence); although I admit we are not consistent in Primes, nor in Base. I think I remember that the guideline was without "s".

src/Primes.jl
rfourquet: I updated again to propose yet another formulation, where I removed (for simplicity) the specification that the factors are of the same type as `n`.

ararslan: I like what you have now. Another option would be to add a docstring to `Factorization`, then just say "returns a ``[`Factorization`](@ref)``," which would tell Documenter to insert a link to the other docstring. Probably not necessary for now though.

rfourquet: Ok let's go then! Also probably good idea to add a docstring to the type itself with a ref, but I guess I prefer to postpone this until I'm more familiar with the doc system.

rfourquet:
Sorry, I used the squash-and-merge from github, but got a redundant message (wanted to force-push to correct that, but I guess we don't do that on public repos?)

PR #39, opened 2017-04-18
ismersenneprime: fix incorrect check test

rfourquet:
The boolean operators (for `check`) where not in the correct order, the precondition that `M` is a Mersenne number was not fully enforced even for `check=true` (need to check that `0` is not in the binary representation of `M`), it was not documented that `p` must be a prime,  and an error was thrown when `M < 7`, which was not documented and not useful. cc @vinctux

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -429,6 +429,7 @@ function ismersenneprime(M::Integer; check::Bool = true)
         M >= 0 && isprime(length(b)) && !('0' in b) ||
             throw(ArgumentError("The argument given is not a valid Mersenne Number (`M = 2^p - 1`)."))
     end
+    M < 7 && return M == 3
     return ll_primecheck(M)
 end
 
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -425,8 +425,8 @@ true
 """
 function ismersenneprime(M::Integer; check::Bool = true)
     if check
-        b = bin(M)
-        M >= 0 && isprime(length(b)) && !('0' in b) ||
+        d = ndigits(M, 2)
+        M >= 0 && isprime(d) && (M >> d == 0) ||
             throw(ArgumentError("The argument given is not a valid Mersenne Number (`M = 2^p - 1`)."))
     end
     M < 7 && return M == 3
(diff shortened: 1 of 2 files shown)


src/Primes.jl
```diff
@@ -422,7 +424,12 @@ true
 ```
 """
 function ismersenneprime(M::Integer; check::Bool = true)
-    (check && isprime(length(bin(M)))) || throw(ArgumentError("The argument given is not a valid Mersenne Number (`M = 2^p - 1`)."))
+    if check
+        b = bin(M)
+        M >= 0 && isprime(length(b)) && !('0' in b) ||
+            throw(ArgumentError("The argument given is not a valid Mersenne Number (`M = 2^p - 1`)."))
```
ararslan: It might be marginally more efficient to avoid processing things as a string and just do:
```julia
if check && M >= 0 && isprime(ndigits(M, 2)) && count_zeros(M) == 0
    throw(ArgumentError(...))
end
```

rfourquet: Yes good idea, except that `count_zeros` counts also leading zeros. Updated accordingly.

rfourquet:
@vinctux Please can you check if this change seems correct to you, as you are the author of the modified function?

src/Primes.jl
```diff
@@ -422,7 +424,12 @@ true
 ```
 """
 function ismersenneprime(M::Integer; check::Bool = true)
-    (check && isprime(length(bin(M)))) || throw(ArgumentError("The argument given is not a valid Mersenne Number (`M = 2^p - 1`)."))
+    if check
+        d = ndigits(M, 2)
+        M >= 0 && isprime(d) && (M >> d == 0) ||
+            throw(ArgumentError("The argument given is not a valid Mersenne Number (`M = 2^p - 1`)."))
+    end
+    M < 7 && return M == 3
```
vinctux: I feel that this optimisation is a bit out of scope of this PR ;) Good thought otherwise :+1: - I'll try to implement various optimisations for LL(R) in the following week.

vinctux:
@rfourquet Thanks for the fixes! I couldn't reply right away, was on vacation. I agree with all of them (also with the `M < 7` case, which is more than just an optimisation, as it fixes a bug ), we shall get this merged.

As in my obsolete comment, I'll try to implement more optimations, which will expand the erroneous `M < 7` case.

PR #41, opened 2017-04-28
add `nextprime`, `prevprime` and `prime`

rfourquet:
So finaly the `nextprime` function!

I still have to add tests, but wanted first to discuss the API: unlike other suggestions, I chose to have `nextprime(n) == n` if `n` is prime. This is what the number theoretic package "pari" does. This is also consistent with `Base.nextpow`. 
Also, there is an optional parameter `i` to get the `i`-th prime number not less than `n`. This is inspired by "sympy". This makes straightforward the implementation of `prime(n)`, which gives the `n`-th prime number.

Also, I set `i::Integer` but that doesn't seem useful here, so I would tend to prefer to set `i::Int` instead (until there is a user request for more generality...)

In a subsequent PR, I plan to optimize those functions by taking advantage of the precomputed `Primes.PRIMES`.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -14,7 +14,7 @@ end
 using Base: BitSigned
 using Base.Checked.checked_neg
 
-export ismersenneprime, isrieselprime
+export ismersenneprime, isrieselprime, nextprime, prevprime, prime
 
 include("factorization.jl")
 
@@ -461,5 +461,123 @@ function ll_primecheck(X::Integer, s::Integer = 4)
     end
     return S == 0
 end
+
+# add_! : "may" mutate the Integer argument (only for BigInt currently)
+
+# modify a BigInt in-place
+function add_!(n::BigInt, x::Int)
+    if x < 0
+        ccall((:__gmpz_sub_ui, :libgmp), Void, (Ptr{BigInt}, Ptr{BigInt}, Culong), &n, &n, -x)
+    else
+        ccall((:__gmpz_add_ui, :libgmp), Void, (Ptr{BigInt}, Ptr{BigInt}, Culong), &n, &n, x)
+    end
+    n
+end
+
+# checked addition, without mutation
+add_!(n::Integer, x::Int) = Base.checked_add(n, oftype(n, x))
+
+"""
+    nextprime(n::Integer, i::Integer=1)
+
+The `i`-th smallest prime not less than `n` (in particular,
+`nextprime(p) == p` if `p` is prime). If `i < 0`, this is equivalent to
+prevprime(n, -i). Note that for `n::BigInt`, the returned number is
+only a pseudo-prime (the function [`isprime`](@ref) is used
+internally). See also [`prevprime`](@ref).
+
+```jldoctest
+julia> nextprime(4)
+5
+
+julia> nextprime(5)
+5
+
+julia> nextprime(4, 2)
+7
+
+julia> nextprime(5, 2)
+7
+```
+"""
+function nextprime(n::Integer, i::Integer=1)
+    i < 0 && return prevprime(n, -i)
+    i == 0 && throw(DomainError())
+    n < 2 && (n = oftype(n, 2))
+    if n == 2
+        if i <= 1
+            return n
+        else
+            n += one(n)
+            i -= 1
+        end
+    else
+        n += iseven(n)
+    end
+    # n can now be safely mutated
+    # @assert isodd(n) && n >= 3
+    while true
+        while !isprime(n)
+            n = add_!(n, 2)
+        end
+        i -= 1
+        i <= 0 && break
+        n = add_!(n, 2)
+    end
+    n
+end
+
+
+"""
+    prevprime(n::Integer, i::Integer=1)
+
+The `i`-th largest prime not greater than `n` (in particular
+`prevprime(p) == p` if `p` is prime). If `i < 0`, this is equivalent to
+`nextprime(n, -i)`. Note that for `n::BigInt`, the returned number is
+only a pseudo-prime (the function [`isprime`](@ref) is used internally). See
+also [`nextprime`](@ref).
+
+```jldoctest
+julia> prevprime(4)
+3
+
+julia> prevprime(5)
+5
+
+julia> prevprime(5, 2)
+3
+```
+"""
+function prevprime(n::Integer, i::Integer=1)
+    i <= 0 && return nextprime(n, -i)
+    n += zero(n) # deep copy of n, which is mutated below
+    while true
+        n < 2 && throw(ArgumentError("There is no prime less than or equal to $n"))
+        while !isprime(n)
+            n = add_!(n, -1)
+        end
+        i -= 1
+        i <= 0 && break
+        n = add_!(n, -1)
+    end
+    n
+end
+
+"""
+    prime{T}(::Type{T}=Int, i::Integer)
+
+The `i`-th prime number.
+
+```jldoctest
+julia> prime(1)
+2
+
+julia> prime(3)
+5
+
+```
+"""
+prime{T<:Integer}(::Type{T}, i::Integer) = i < 0 ? throw(DomainError()) : nextprime(T(2), i)
+prime(i::Integer) = prime(Int, i)
 
 end # module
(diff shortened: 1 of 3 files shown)


src/Primes.jl
```diff
@@ -474,4 +474,121 @@ function ll_primecheck(X::Integer, s::Integer = 4)
     return S == 0
 end
 
+
+function add!!(n::BigInt, x::Int)
+    if x < 0
+        ccall((:__gmpz_sub_ui, :libgmp), Void, (Ptr{BigInt}, Ptr{BigInt}, Culong), &n, &n, -x)
+    else
+        ccall((:__gmpz_add_ui, :libgmp), Void, (Ptr{BigInt}, Ptr{BigInt}, Culong), &n, &n, x)
```
ararslan: You could avoid the branch here by using `abs(x)`

src/Primes.jl
```diff
@@ -474,4 +474,121 @@ function ll_primecheck(X::Integer, s::Integer = 4)
...
+julia> prevprime(5)
+5
+
+julia> prevprime(5, 2)
+3
+```
+"""
+function prevprime(n::Integer, i::Integer=1)
+    i <= 0 && return nextprime(n, -i)
+    n = n+zero(n) # deep copy of n
```
ararslan: `n += zero(n)`? Also I'm not sure what you mean by a deep copy of `n`; I didn't think integer variables could be aliased.

src/Primes.jl
```diff
@@ -474,4 +474,121 @@ function ll_primecheck(X::Integer, s::Integer = 4)
...
+```jldoctest
+julia> prime(1)
+2
+
+julia> prime(3)
+5
+
+```
+"""
+prime(T::Type, i::Integer) = i < 0 ? throw(DomainError()) : nextprime(T(2), i)
```
ararslan: Presumably we want to restrict `T` to subtypes of integers, so this could be
```julia
prime{T<:Integer}(::Type{T}, i::Integer) = i < 0 ? throw(DomainError()) : nextprime(T(2), i)
```

src/Primes.jl
rfourquet: But it's not the same library function which is called. Will check though whether it's possible to put a branch within the `ccall`.

src/Primes.jl
rfourquet: Sure. Deep copy: I mean that later in this routine `n` is mutated for `BigInt`, but this should not mutate original `n` which is passed as argument, hence a deep copy is needed, i.e. which copies the internal (no sharing). I tried to give `n` a different name after the copy, but it complicates slighly the function.

src/Primes.jl
rfourquet: Yes, I was relying on the fact that the restriction would anyway happen in the call to `nextprime`, but you are probably right, also for the docstring.

src/Primes.jl
rfourquet: Nope, `ccall` expects a constant expression.

src/Primes.jl
rfourquet: I confused nextprime and prevprime, it would be easy to give another name for `n` in `prevprime` (eg. `m = n+zero(n)`). I will add a test which checks this property on which we rely here.

rfourquet:
Also, in pari `prevprime` is spelled `precprime`, which name is better?

ararslan:
> which name is better?

`prevprime`, IMO. What's `prec` short for?

src/Primes.jl
ararslan: > But it's not the same library function which is called.

Ohhhhhhh right, sorry, I missed that. The fact that they're the same length fooled my eyes.

src/Primes.jl
ararslan: > this should not mutate original `n` which is passed as argument

Right, but my understanding is that integer variables _can't_ be mutated. Is that not true?

ararslan: Apparently `BigInt`s are mutable, but `+=` does not mutate them.

tkelman:
preceding?

tkelman:
but could be ambiguous with precision etc, so I agree prev is more idiomatic in julia (c.f. findprev)

rfourquet:
Ok will keep `prevprime` (also more consistent with `prevpow`). I will assume the API is ok as no objection came, and will write corresponding tests.

src/Primes.jl
rfourquet: The API is that integers are value types, immutables, but in practice `BigInt` are a `mutable struct`, of which we can take advantage for optimization, as long as this is not visible to the user. So it's acceptable to mutate a `BigInt` only if the object is under control, e.g. a local variable. When doing `n = n+0`, the addition creates a fresh new `BigInt`, not aliasing the function argument, and hence can be safely mutated. I test this property so that if this case becomes specially handled in the addition code to not allocate a new `BigInt`, we can detect it.

rfourquet:
For me, the main question is whether we keep `nextprime(p) == p` when `p` is prime. Will keep this open few more days for giving time for more comments.

rfourquet:
@simonbyrne , @pabloferz , @mschauer , do you like to comment on this change before we merge?

rfourquet:
I plan to merge tomorrow if no more discussions come (in particular concerning the API...)

src/Primes.jl
```diff
@@ -462,4 +462,121 @@ function ll_primecheck(X::Integer, s::Integer = 4)
...
+        ccall((:__gmpz_sub_ui, :libgmp), Void, (Ptr{BigInt}, Ptr{BigInt}, Culong), &n, &n, -x)
+    else
+        ccall((:__gmpz_add_ui, :libgmp), Void, (Ptr{BigInt}, Ptr{BigInt}, Culong), &n, &n, x)
+    end
+    n
+end
+
+add!!(n::Integer, x::Int) = Base.checked_add(n, oftype(n, x))
+
+macro add!!(n, x)
```
rfourquet: Simply my personal preference, so that updating in-place for `BigInt` costs as less syntactic noise as possible...
`@add!! n 1` vs `n = add!!(n, 1)`. If one of you prefers me to use the function instead, I will do!
BTW, are you fine with two `!!` meaning here "mutating in some cases" ?

pabloferz: IMHO I don't think that in this particular case we gain enough with the macro for justifying it, so I'd prefer just the function.

w.r.t the `!!` I like better something like `add!_` (for my `!!` gives the impression of being stronger that just `!`).

rfourquet: `!!` was proposed by someone I think (maybe @TotalVerb ?), but I can't find the reference now... Let's see what @ararslan or someone else thinks about it.

ararslan: I'm not a huge fan of the `!!` since it seems like it's shouting at me, and I also prefer a function to a macro. When reviewing this I didn't think much of it since it isn't exported.

It's actually `Base.checked_add` that's doing the possible mutating, right? If that's the case, perhaps it could just be `add` with a note in the docstring regarding potential mutation? I'd also be fine with simply `add!`. Though as long as it isn't exported, I don't care all that much.

rfourquet: I will remove the macro then no problem. `add!!` has basically two methods, one for `BigInt` (which is mutating, for efficency) and one for generic `Integer` (which isn't mutating, e.g. `Int` cannot be mutated, but which does check for overflow). So the `!!` meant "can be mutating, but not always". I would be fine with `add!` also, and indeed it's not exported so I don't care too much about the name.

rfourquet:
I merged from the command line, and for some reason this PR didn't show as merged. So I tried to push my updated branch, without better result, and endend up push the new master onto this branch, which finally appear merged now, but it's not clean, does someone know what to do? Maybe just repush the old branch and close manually? Sorry for messing this up, I tried a variation of the instructions on how to merge on the command line, but I didn't know the way github really works...

ararslan:
cc @tkelman, master of all things (which includes git) ^

stevengj:
See this [complaint on the mailing list](https://discourse.julialang.org/t/the-prevprime-and-nextprime-functions/4163) that the behavior `nextprime(p) == p` is inconsistent with Mathematica and Maple.  It is also apparently inconsistent with SymPy.    I don't have strong feelings on the subject, myself.

PR #42, opened 2017-04-28
add `prodfactors` and `radical` functions

rfourquet:
EDIT TITLE: renamed `factorback` -> `prodfactors` (not updated below).
Add `factorback` such that `factorback(Container, factor(n)) == n` (cf. docstrings for details). 
The name ("factorback" comes from the "pari" package) can of course be bikeshed, cf. https://github.com/JuliaLang/julia/issues/12153. 
I like also the name `factorprod` (but there was an objection in the linked issue). 
I also added the function `radical` as it was natural to do so, but didn't export it (yet).

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -4,6 +4,8 @@
 
 ```@docs
 Primes.factor
+Primes.prodfactors
+Primes.radical
 ```
 
 ## Generating prime numbers
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -14,7 +14,7 @@ end
 using Base: BitSigned
 using Base.Checked.checked_neg
 
-export ismersenneprime, isrieselprime, nextprime, prevprime, prime
+export ismersenneprime, isrieselprime, nextprime, prevprime, prime, prodfactors, radical
 
 include("factorization.jl")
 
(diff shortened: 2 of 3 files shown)


src/Primes.jl
```diff
@@ -358,6 +358,44 @@ factor{T<:Integer}(::Type{Vector{T}}, n::T) =
...
+to `Base.prod`.
+
+```jldoctest
+julia> factorback(factor(100))
+100
+```
+"""
+function factorback end
+
+factorback(factors::Associative) = prod(p^i for (p, i) in factors)
```
ararslan: Julia 0.4 has array comprehensions but not `Generator`s, which is causing the test failures on 0.4. I think we should just drop 0.4 support at this point. I can submit a PR for that later.

ararslan:
> I also added the function radical as it was natural to do so, but didn't export it (yet).

Might as well export it, IMO, or at least add it to the docs. Seems pretty cool.

ararslan:
I'm not 100% sold on the name `factorback`, as I don't find it immediately descriptive, but I'm also not really opposed to it. A term that comes to mind is "expand," as the inverse of factorization is expansion when talking about polynomials, but I don't know whether that term makes sense for integer factorizations.

src/Primes.jl
rfourquet: Oh I keep forgetting about that!

rfourquet:
I actually also like `expand`, and thought about it but was drawn away by the fact that it has a totally different meaning in Base. But it is also a shame that this name (with mathematical meaning) would be reserved for a technical functionality of julia code. Considering opening a Julia issue for freeing that name...

rfourquet:
Ok, exported `radical`, removed the generator expression and added a test in the hope to have coveralls happier.

ararslan:
Even if we could reclaim `expand` from Base, it would be a while before we could do so. First it would have to be deprecated there, and it's too late to add new deprecations to 0.6, so it would have to be deprecated in the next release and removed after. So the soonest we could use it would be 1.0, and we still wouldn't be able to use if then if we wanted to maintain support for older Julia versions.

Now, if we wanted to support only going from a `Primes.Factorization` to an integer, we could overload `Base.expand` for that, since our factorization type is not in Base and `expand` currently only takes an `Expr` as an argument. But it's type piracy if we wanted to use anything more general for reconstructing the integer, e.g. any `Associative`.

We could do something like `expandfactors`. It's a little long, but it's descriptive.

src/Primes.jl
```diff
@@ -346,6 +346,44 @@ factor{T<:Integer}(::Type{Vector{T}}, n::T) =
...
+to `Base.prod`.
+
+```jldoctest
+julia> factorback(factor(100))
+100
+```
+"""
+function factorback end
+
+factorback(factors::Associative) = prod([p^i for (p, i) in factors])
```
ararslan: Support for 0.4 is now a thing of the past, so this can be the bare generator expression again if you'd prefer. I did a benchmark to see whether having it in an array comprehension versus a generator made a difference and it doesn't for the most part, so changing isn't really necessary, just thought I'd give you a heads up.

docs/src/api.md
```diff
@@ -4,6 +4,7 @@
 
 ```@docs
 Primes.factor
+Primes.factorback
```
ararslan: Could you add `radical` here too? (Or wherever the appropriate place in the docs is)

rfourquet:
Indeed `Base.expand` would be type piracy for anything other than `Primes.Factorization`, for which I already proposed `prod` here. I realize that indeed a general `Associative`, the name should contain `factor`. As `prod` seems good (to me) for `Factorization`, I find `factor(s)prod` or `prodfactors` more consistent. Though `expandfactors` sounds good too. Will keep open few more days to allow more people to comment on a prefered name.

docs/src/api.md
rfourquet: Good catch.

rfourquet:
Someone likes to feed the bikeshed here?!

rfourquet:
Instead of `factorback`, what about having a non-exported `Primes.prod`? `Base.prod` would work for `Primes.factorization`, and `Primes.prod` would have to be used for other containers...

ararslan:
Hm, I'm not really a fan of shadowing Base functions like that. I do still like `expandfactors` but if you're not into that, I think `factorback` is fine.

rfourquet:
Let's forget about the shadowing (of which I'm not a big fan either), I prefer `expandfactors` to `factorback`, but I think I prefer even more `prodfactors` because `prod` is already used for `Factorization`. Opinions @pabloferz @simonbyrne ?

ararslan:
Ah yeah, I like `prodfactors`.

Issue #49, opened 2017-06-21
Factor is terribly slow


enedil:
Hello.
I believe that after finding every consecutive prime factor, performing `isprime` on resultant is harmful for performance of such function. Consider such snippet:
```
number = reduce(*, 1, Primes.PRIMES)
tic(); factor(number); toc()
# already running for 15 minutes, still didn't halt
```
Compare it with equivalent Python:
```
from operator import mul
from sympy import primerange, factorint
number = reduce(mul, primerange(1, 2**16), 1)
%timeit factorint(number)
# 1 loops, best of 3: 16.8 s per loop
```
Considering the fact that `sympy` is written in pure Python, this is pathetic.

Profiling Julia version brings such result (this time taking the product of first 3000 primes):
```
WARNING: The profile data buffer is full; profiling probably terminated
before your program finished. To profile for longer runs, call Profile.init
with a larger buffer and/or larger delay.
18816 ./event.jl:68; (::Base.REPL.##3#4{Base.REPL.REPLBackend})()
 18816 ./REPL.jl:95; macro expansion
  18816 ./REPL.jl:64; eval_user_input(::Any, ::Base.REPL.REPLBackend)
   18816 ./<missing>:?; anonymous
    18816 ./profile.jl:16; macro expansion;
     18816 /home/enedil/.julia/v0.5/Primes/src/Primes.jl:312; factor(::BigInt)
      1     ./dict.jl:701; factor!(::BigInt, ::Dict{BigInt,Int64})
      11    /home/enedil/.julia/v0.5/Primes/src/Primes.jl:267; factor!(::BigInt, ::Dict{BigInt,Int64})
       11 ./gmp.jl:110; convert
      3     /home/enedil/.julia/v0.5/Primes/src/Primes.jl:268; factor!(::BigInt, ::Dict{BigInt,Int64})
       1 ./gmp.jl:255; rem
       2 ./gmp.jl:256; rem
      2     /home/enedil/.julia/v0.5/Primes/src/Primes.jl:269; factor!(::BigInt, ::Dict{BigInt,Int64})
       1 ./dict.jl:644; setindex!(::Dict{BigInt,Int64}, ::Int64, ::BigInt)
      3     /home/enedil/.julia/v0.5/Primes/src/Primes.jl:270; factor!(::BigInt, ::Dict{BigInt,Int64})
       3 ./gmp.jl:256; div
[...]

PR #50, opened 2017-06-21
Replace isprime

enedil:
Replace isprime with weaker, but orders of magnitude faster test. Better to finish testing primes up to 2^16, then performing Miller-Rabin as many times as different prime factors.

Related issue: https://github.com/JuliaMath/Primes.jl/issues/49

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -267,6 +267,7 @@ function factor!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
             p^2 >= n && (h[n] = 1; return h)
         end
     end
+    isprime(n) && (h[n] = 1; return h)
     T <: BigInt || widemul(n - 1, n - 1) ≤ typemax(n) ? pollardfactors!(n, h) : pollardfactors!(widen(n), h)
 end
 
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -264,7 +264,7 @@ function factor!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
                 n = div(n, p)
             end
             n == 1 && return h
-            isprime(n) && (h[n] = 1; return h)
+            p^2 >= n && (h[n] = 1; return h)
         end
     end
     T <: BigInt || widemul(n - 1, n - 1) ≤ typemax(n) ? pollardfactors!(n, h) : pollardfactors!(widen(n), h)


src/Primes.jl
```diff
@@ -264,9 +264,10 @@ function factor!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
                 n = div(n, p)
             end
             n == 1 && return h
-            isprime(n) && (h[n] = 1; return h)
+            p^2 >= n && (h[n] = 1; return h)
```
rfourquet: If `p^2 >= n`, it means that `n` is the factor of primes less than or equal to `p`: hence `n` should have already been factored out by that time, and the condition `n == 1` already met!

rfourquet:
So you are proposing a different strategy: can you show some timings, both what speed-up or slow-down can be expected?

src/Primes.jl
rfourquet: A more sensible change would be to test `PRIMES[end]^2 >= n` as a condition to break out of the loop.

enedil: @rfourquet no, that's exactly what I meant. Either n is 1, or n is prime by this moment.

enedil:
@rfourquet i'll reply when I'll get to my notebook

enedil:
@rfourquet 
Ok - some tests:
```
test_numbers = [#Primes.PRIMES,
                [921236161, 931235651, 941236273, 951236179, 901236131, 961236149, 911236121],
                [2, 3, 5, 13, 17, 61, 623341, 252097800623, 252097800629, 252097800667, 2760727302517],
                [63721832374, 13642987325257, 13789015084387, 25092045343291, 18748357090643, 19019197924633, 1927861641436, 115215177288657],
                [2, 23]*1000000000]
for n in test_numbers
    num = reduce(*, big(1), n)
    tic(); factor(num); toc()
end
```

And results:
```
~/c/g/P/src ❯❯❯ julia old.jl 
elapsed time: 0.892513902 seconds
elapsed time: 5.628045027 seconds
elapsed time: 40.111286483 seconds
elapsed time: 5.1307e-5 seconds
~/c/g/P/src ❯❯❯ julia new.jl 
elapsed time: 0.215181272 seconds
elapsed time: 1.384578026 seconds
elapsed time: 43.26424359 seconds
elapsed time: 4.0111e-5 seconds
```

Every factor in the lists is prime. I commented out `Primes.PRIMES` array, so that I can ever finish factoring in a reasonable time. As seen, old implementation is rather generally slower (although in some cases another test run might change the output).

As for `Primes.PRIMES` array, time for `new.jl` is:

    elapsed time: 0.553304032 seconds

This is considerably faster that at least 15 minutes.

src/Primes.jl
rfourquet: But can you give me an exampleof initial value of `n` such that `p^2 >= n > 1` at this point?

rfourquet:
OK thanks for benchmarking. What about a compromise: adding an `isprime` test every 10 or 100 iteration?

src/Primes.jl
enedil: One example is 63. After factoring out 3*3, we are left with 7, which satisfies conditions : 1< 7 < 3^2. As I said, it is a prime. Also, equal sign is here redundant, but the point still holds. 

Let's put it this way:
You might want to check PRIMES[end]^2 >= n. Sure, if it is met, then n is at that point either prime or 1. But we could test something similar - p^2 >= n (p being the rithg now excluded factor). If this condition is met, then suppose n=a*b with a, b proper divisors of n. Bot a and b are at this point greater then p. So n=a*b>p*p>=n, a contradiction. So n has only trivial factorisation, hence it's a prime. 

But it is a considerably stronger test! For if n falls into the latter, it falls also into the former, but nor in converse. That is why I propose such solution.

enedil:
The point is - numbers with large number of prime factors and numbers with very little prime factors, both are rather rare. 

You may be right. In fact, Primes.PRIMES table is about 6300 long. Doing Rabin-Miller after every iteration is not necessary, but if it was only 63 times, maybe it would make sense? I'd have to do timings.

src/Primes.jl
rfourquet: Yes of course, thank you and sorry, I should have analysed your change more carefully!

rfourquet:
When there are very few factors from `PRIMES`, maybe the test `PRIMES[end]^2 > n` can be used to break out of the loop early (meaning that either `n` is prime, or at least has factors greater than `PRIMES[end]`) ?

enedil:
That is not true. What if `n = 2039^3 * 3`? `2039` is prime. After ruling out `3`, we're left with `2039^3 = 8477185319 > 4293001441 = 65521^2 = PRIMES[end]^2`.

enedil:
Maybe there should be an optional argument which is how often make tests and default it to about 100? And make docs that say what is the specifics of such argument? And furthermore, don't perform `isprime`, which tests for first 9 primes, maybe `isprime` should be splitted so every phase has its own function. How do you see such design? 

Also, excuse me if my code is poor, this pr is literally first 3 lines I ever wrote in Julia.

rfourquet:
> Maybe there should be an optional argument which is how often make tests 

I would say it's a bit premature to offer an optional argument for that. And if someone has a wish that this option is available, meaning she has an insight how to optimize this value, maybe would be good then to include this heuristic in this package! 

> maybe isprime should be splitted so every phase has its own function. How do you see such design?

It seems like a good idea, I thought about that too. Whether you want to implement that in this PR or in another one is your call!

enedil:
I'd better if we finished this PR (separation of concerns). Then I will start with clean state. And as far as optional arg is concerned, I found that in very most cases, the less prime tests, the better. So I'd stick with about 200 probably.

Btw, do you have any idea why "AppVeyor build failed"? Logs mention reducing over empty collection, I cannot locate where this function could yield empty Dict.

rfourquet:
> So I'd stick with about 200 probably.

OK.

> Btw, do you have any idea why "AppVeyor build failed"? Logs mention reducing over empty collection, I cannot locate where this function could yield empty Dict.

Try `factor(Dict, 1)`. I will open a PR for that.

PR #51, opened 2017-06-24
fix prodfactor(factor(1)) (was an error)

rfourquet:


src/Primes.jl
```diff
@@ -362,7 +362,7 @@ julia> prodfactors(factor(100))
 """
 function prodfactors end
 
-prodfactors(factors::Associative) = prod(p^i for (p, i) in factors)
+prodfactors{K}(factors::Associative{K}) = isempty(factors) ? one(K) : prod(p^i for (p, i) in factors)
```
ararslan: I can't think of an example where this would matter, but it might be worth doing a promotion on `K` before returning the multiplicative unit of its type to ensure type stability. Unless `T(x1)^Int(y1) * T(x2)^Int(y2) * ...` always has type `T`, in which case this is fine.

rfourquet: For all integer types in Base, this is type-stable... A conversion to `K` doesn't seem necessary, but would also not be a problem.

ararslan: As long as it's type stable for existing types that's good enough for me. :)

rfourquet: For me too! Let's review again if there are complains. But even in some cases (when the argument `factors` doesn't come from the function `factor`, it could be that the produce overflows, unless the type `K` promotes to a wider type for `^`. In this case, it may be worth allowing non-type-stability? (My example is a bit far-stretched, I agree ;-) )

Issue #46, opened 2017-05-24
Performance regression in isprimes


KristofferC:
Moved from https://github.com/JuliaLang/julia/issues/16780

The function `isprimes`  of #16349, or more precisely of https://github.com/JuliaMath/Primes.jl/pull/9 with type hint for the `witnesses` observed a significant drop in speed recently, see the discussion at  https://github.com/JuliaMath/Primes.jl/pull/9. When I was writing the comment https://github.com/JuliaLang/julia/pull/16349#discussion_r63227124 (using a slightly outdated master) no allocation took place in the inner loops at all. This seems to have changed recently as well and allocation takes place.

oscardssmith:
I believe this has been fixed by now.

PR #53, opened 2017-06-24
Better algorithm for factoring + refactoring of isprime

enedil:
Here I implement proposed changes from previous PR (so I gonna close it), with better structured code.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -281,7 +281,7 @@ function factor!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
             n == 1 && return h
             p^2 > n && (h[n] = 1; return h)
             counter += 1
-            if counter % 75 == 0 || i - last_test > 300
+            if counter % 75 == 0 || i - last_test > 700
                 last_test = i
                 _miller_rabin(n) && (h[n] = 1; return h)
             end
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -293,7 +293,7 @@ end
 
 
 """
-factor(n::Integer) -> Primes.Factorization
+    factor(n::Integer) -> Primes.Factorization
 
 Compute the prime factorization of an integer `n`. The returned
 object, of type `Factorization`, is an associative container whose
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -133,7 +133,7 @@ primes(n::Int) = primes(1, n)
 
 const PRIMES = primes(2^16)
 
-function _miller_rabin(n::Unsigned)
+function _miller_rabin(n::Integer)
     s = trailing_zeros(n - 1)
     d = (n - 1) >> s
     for a in witnesses(n)::Tuple{Vararg{Int}}
@@ -150,10 +150,10 @@ function _miller_rabin(n::Unsigned)
 end
 
 function _trial_division(n::Integer, limit::Integer)
-    for p in PRIMES
+    for p in (2, 3, 5, 7, 11, 13, 17, 19, 23, 29)
         p > limit && return Nullable{Bool}() 
-        p^2 > n && return Nullable{Bool}(true)
-        n % p == 0 && return Nullable{Bool}(false)
+        p^2 > n && return Nullable(true)
+        n % p == 0 && return Nullable(false)
     end
 end
 
@@ -176,8 +176,7 @@ function isprime(n::Integer)
     t = _trial_division(n, 25)
     !isnull(t) && return get(t)
     
-    n = Unsigned(n)
-    return _miller_rabin(n)
+    return _miller_rabin(Unsigned(n))
 end
 
 """
@@ -282,7 +281,7 @@ function factor!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
             n == 1 && return h
             p^2 > n && (h[n] = 1; return h)
             counter += 1
-            if counter % 75 == 0 || i - last_test > 700
+            if counter & 0x3f == 0 || i - last_test > 700
                 last_test = i
                 _miller_rabin(n) && (h[n] = 1; return h)
             end
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -133,9 +133,9 @@ primes(n::Int) = primes(1, n)
 
 const PRIMES = primes(2^16)
 
-function _miller_rabin(n::Integer)
+function _miller_rabin(n::Unsigned)
     s = trailing_zeros(n - 1)
-    d = (n - 1) >>> s
+    d = (n - 1) >> s
     for a in witnesses(n)::Tuple{Vararg{Int}}
         x = powermod(a, d, n)
         x == 1 && continue
@@ -176,6 +176,7 @@ function isprime(n::Integer)
     t = _trial_division(n, 25)
     !isnull(t) && return get(t)
     
+    n = Unsigned(n)
     return _miller_rabin(n)
 end
 
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -136,7 +136,7 @@ const PRIMES = primes(2^16)
 function _miller_rabin(n::Integer)
     s = trailing_zeros(n - 1)
     d = (n - 1) >> s
-    for a in witnesses(n)::Tuple{Vararg{Int}}
+    for a in witnesses(n)
         x = powermod(a, d, n)
         x == 1 && continue
         t = s
@@ -176,7 +176,7 @@ function isprime(n::Integer)
     t = _trial_division(n, 25)
     !isnull(t) && return get(t)
     
-    return _miller_rabin(Unsigned(n))
+    return _miller_rabin(n)
 end
 
 """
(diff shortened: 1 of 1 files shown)


src/Primes.jl
```diff
@@ -147,23 +171,12 @@ function isprime(n::Integer)
...
-            (t -= 1) ≤ 0 && return false
-            x = oftype(n, widemul(x, x) % n)
-            x == 1 && return false
-        end
-    end
-    return true
+    n < 0 && return false
+
+    t = Nullable{Bool}(_trial_division(n, 25))
+    !isnull(t) && return get(t)
```
ararslan: I'm confused by the use of `Nullable` here. Since `_trial_division` returns a `Bool`, why not just use that directly rather than wrapping and unwrapping the `Nullable`?

src/Primes.jl
```diff
@@ -264,15 +279,21 @@ function factor!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
...
         end
     end
+    isprime(n) && (h[n] = 1; return h)
     T <: BigInt || widemul(n - 1, n - 1) ≤ typemax(n) ? pollardfactors!(n, h) : pollardfactors!(widen(n), h)
 end
 
 
 """
-    factor(n::Integer) -> Primes.Factorization
+factor(n::Integer) -> Primes.Factorization
```
ararslan: The indent here should be maintained

src/Primes.jl
enedil: No. Notice that if `_trial_division` doesn't find any factor in range, but isn't assured about primality of a number, it returns null. Ex. `_trial_division(23, 4)` will be null, because perhaps `23` has some prime divisor, but check is made only up to `4`.

src/Primes.jl
enedil: @ararslan fixed

src/Primes.jl
ararslan: Ah, but it's not null, it's `nothing`. It would be better to explicitly return a `Nullable` from `_trial_division`.

src/Primes.jl
ararslan: Thanks

src/Primes.jl
enedil: Sure, these contributions are my very first lines of Juila, I still gotta learn it. Thanks

ararslan: Oh wow, excellent first Julia!

src/Primes.jl
```diff
@@ -133,6 +133,30 @@ primes(n::Int) = primes(1, n)
 
 const PRIMES = primes(2^16)
 
+function _miller_rabin(n::Integer)
+    s = trailing_zeros(n - 1)
+    d = (n - 1) >>> s
```
musm: I would double check the performance. Sometimes the `>>>` can't tell that `s` is an unsigned integer and hence the bit shift is slow.

musm: I meant you should put `Unsigned(s)` and just use `>>`  since `trailing_zeros` always gives a positive number, but the shift operator won't be able to infer it is an unsigned integer without being explicit

musm:
If I understand `_miller_rabin` is an internal function, leaving it as `::Integer` is just as good since the outer `isprime` function has the `n < 2` check

enedil:
@musm reverting because a) your comment has a point b) I would have to vhange tests otherwise.

enedil:
@musm reverting because a) your comment has a point b) I would have to change tests otherwise.

enedil:
Forgive me those duplicated comments

ararslan:
No problem

src/Primes.jl
```diff
@@ -133,6 +133,30 @@ primes(n::Int) = primes(1, n)
 
 const PRIMES = primes(2^16)
 
+function _miller_rabin(n::Unsigned)
```
rfourquet: So the tests seem to fail because of this `::Unsigned` signature. I'm in favor to keep `Integer`, it would allow to keep a `BigInt` and I don't understand what `Unsigned` solves.

src/Primes.jl
```diff
@@ -133,6 +133,30 @@ primes(n::Int) = primes(1, n)
...
+            (t -= 1) ≤ 0 && return false
+            x = oftype(n, widemul(x, x) % n)
+            x == 1 && return false
+        end
+    end
+    return true
+end
+
+function _trial_division(n::Integer, limit::Integer)
+    for p in PRIMES
```
rfourquet: @pabloferz had removed the use of `PRIMES` in #9 because of performances considerations (maybe `PRIMES` causes cache misses). Unless you are willing to do some benchmarks comparing using `PRIMES` vs. `(2, 3, ..., 23)` as before, I would favor keeping the the latter as before (and removing the `limit` arg).

src/Primes.jl
```diff
@@ -133,6 +133,30 @@ primes(n::Int) = primes(1, n)
...
+        p > limit && return Nullable{Bool}() 
+        p^2 > n && return Nullable{Bool}(true)
```
rfourquet: I think usually we'd rather write `Nullable(true)` directly, the type being inferred easily here.

src/Primes.jl
```diff
@@ -147,23 +171,13 @@ function isprime(n::Integer)
...
-            x == 1 && return false
-        end
-    end
-    return true
+    n < 2 && return false
+
+    t = _trial_division(n, 25)
+    !isnull(t) && return get(t)
+    
+    n = Unsigned(n)
```
rfourquet: I confirm I'd rather not use `Unsigned` here. At the very least, call `_miller_rabin(Unsigned(n))` instead of `n=Unsigned`.

src/Primes.jl
```diff
@@ -264,9 +280,15 @@ function factor!{T<:Integer,K<:Integer}(n::T, h::Associative{K,Int})
                 n = div(n, p)
             end
             n == 1 && return h
-            isprime(n) && (h[n] = 1; return h)
+            p^2 > n && (h[n] = 1; return h)
+            counter += 1
+            if counter % 75 == 0 || i - last_test > 700
```
rfourquet: I don't know if this matters here, but it may be cheaper to use a modulo via a mask, e.g. `counter & 0x3f` (equivalent to `counter % 64).

rfourquet:
I think you could have kept working on your previous PR here (i.e. you could have put this work in the other PR without opening a new one, but not a problem though).

src/Primes.jl
```diff
@@ -133,6 +133,30 @@ primes(n::Int) = primes(1, n)
 
 const PRIMES = primes(2^16)
 
+function _miller_rabin(n::Integer)
+    s = trailing_zeros(n - 1)
+    d = (n - 1) >> s
+    for a in witnesses(n)::Tuple{Vararg{Int}}
```
pabloferz: IIRC from some comments in another issue I believe the type assertion here might affect performance, so I'd remove it.

pabloferz:
So far it looks as a good first contribution :+1:, I'll look into more carefully. In any case it would be nice to have some performance comparisons between the old and new implementations.

enedil:
@pabloferz in general it's hard to benchmark when differences are less than 5%, while Hyper Threading may cause random gain or loss of 15%. Results are pretty similar in random situation (I generated a bit of big numbers and I get at most 3% worse results with old implementation). The difference is in special cases, i.e. when there are a lot of small factors.

src/Primes.jl
enedil: What do you mean? `Tuple{Vararg{Int}}` or what?

pabloferz: Yeah, the `::Tuple{Vararg{Int}}`

rfourquet: @pabloferz you added this type annotation in #9 for improved performances, no?

pabloferz: Yes, but there seems to have been a regression for some reason. See https://github.com/JuliaMath/Primes.jl/issues/46

rfourquet: Ah OK, I had missed that one, and can't understand why it's less efficient that way... But at least while comparing performances between master and this PR, the annotation should stay (to not mix improvements).

pabloferz: Yeah, you're right, that's another issue and should be addressed separately. We should keep the type annotation for now.

rfourquet:
So in some cases, the improvements are very nice (like 60x speedup for the product of `primes(2^10, 2^11)` ) ! But in other cases, there is also a huge regression, like when factoring `1027253819`, where I observe a 17x slowdown... is it possible to tune differently your heuristics to mitigate a bit more the regressions?

src/Primes.jl
```diff
@@ -147,23 +171,12 @@ function isprime(n::Integer)
...
-            x == 1 && return false
-        end
-    end
-    return true
+    n < 2 && return false
+
+    t = _trial_division(n, 25)
+    !isnull(t) && return get(t)
+    
+    return _miller_rabin(Unsigned(n))
```
musm: based on my testing the `Unsigned` seems unnecessary here

enedil:
Ok. Apparently, there is some way (which makes general random numbers perform a bit slower, but which however make those edge cases be a lot faster.

rfourquet:
OK, indeed factoring `1027253819` is now only 3 times slower than master.

I would like to suggest to split your PR: keep this one only for refactoring `isprime`, and then re-open #50 to modify `factor` to take advantage of the refactoring. If this was only a refactoring, I would have no problem, but you modify also slightly the `isprime` algorithm, so it's hard to track the performances implications of two separate changes in a single PR.

PR #54, opened 2017-06-25
make isprime(::BigInt) use deterministic algo when possible

rfourquet:
When a `BigInt` is small enough (`n<=2^64` currently), I don't see why it shouldn't use the deterministic algorithm, in particular as it is usually not much slower (and can be faster sometimes). 
I also convert signed types to unsigned as it seems to be usually faster (but I tested only on 1 machine!)

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -135,12 +135,23 @@ const PRIMES = primes(2^16)
 
 """
     isprime(n::Integer) -> Bool
+    isprime(n::Integer, reps) -> Bool
 
 Returns `true` if `n` is prime, and `false` otherwise.
+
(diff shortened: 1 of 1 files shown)


rfourquet:
@oscardssmith I have a hard time finding time to follow the recent discussions in this repo, I see you have multiple open PRs, which one would you like to be reviewed first? And I totally forgot about this open PR you closed here, is it not valid anymore, or is it superseded by another PR?

oscardssmith:
I probably should have left a message before closing this, but TLDR is that GMP does this for us. From my brief benchmarking, just calling to GMP is about the same speed as converting first.

Can you review https://github.com/JuliaMath/Primes.jl/pull/102 first? It's relatively simple, but a massive speedup.

PR #56, opened 2017-06-26
add nextprimes/prevprimes iterators (fix #55)

### LINKED ISSUE #55: Prime iterator
Implement a prime iterator. This was discussed in #8 . There was also a related question on stackoverflow indicating interest in such a thing, https://stackoverflow.com/q/37142821/2556061

rfourquet:
This adds a `nextprimes(n)` iterator to iterate over all primes greater than `n`. This comes also with an "array" version `nextprimes(n, k)` containing `k` elements (I saw myself too often using the horribly inefficient `[nextprime(n, i) for i in 1:k]` ;-p)
Similarly for `prevprimes`.
Of course, the names can be bikeshed :)

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -671,7 +671,7 @@ prime(::Type{T}, i::Integer) where {T<:Integer} = i < 0 ? throw(DomainError(i)) 
 prime(i::Integer) = prime(Int, i)
 
 
-struct NextPrimes{T<:Integer}
+immutable NextPrimes{T<:Integer}
     start::T
 end
 
@@ -719,7 +719,7 @@ julia> nextprimes(10, 3)
 nextprimes{T<:Integer}(start::T, n::Integer) =
     iterate(x->nextprime(add(x, 1)), nextprime(start), n)
 
-struct PrevPrimes{T<:Integer}
+immutable PrevPrimes{T<:Integer}
     start::T
 end
 
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -732,7 +732,7 @@ iterate(np::PrevPrimes, state=np.start) =
     end
 
 IteratorSize(::Type{<:PrevPrimes}) = Base.SizeUnknown()
-Iteratoreltype(::Type{<:PrevPrimes}) = Base.HasEltype()
+IteratorEltype(::Type{<:PrevPrimes}) = Base.HasEltype()
 
 eltype(::Type{PrevPrimes{T}}) where {T} = T
 
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -682,7 +682,7 @@ Base.done(np::NextPrimes, state) = false
 Base.iteratorsize(::Type{<:NextPrimes}) = Base.IsInfinite()
 Base.iteratoreltype(::Type{<:NextPrimes}) = Base.HasEltype() # default
 
-Base.eltype{T}(::Type{NextPrimes{T}}) = T
+Base.eltype(::Type{NextPrimes{T}}) where {T} = T
 
 """
     nextprimes(start::Integer)
@@ -698,7 +698,7 @@ nextprimes(start::Integer) = NextPrimes(start)
 Returns an iterator over all primes, with type `T`.
 Equivalent to `nextprimes(T(1))`.
 """
-nextprimes{T<:Integer}(::Type{T}) = nextprimes(one(T))
+nextprimes(::Type{T}) where {T<:Integer} = nextprimes(one(T))
 nextprimes() = nextprimes(Int)
 
 """
@@ -716,7 +716,7 @@ julia> nextprimes(10, 3)
  17
 ```
 """
-nextprimes{T<:Integer}(start::T, n::Integer) = collect(T, take(nextprimes(start), n))
+nextprimes(start::T, n::Integer) where {T<:Integer} = collect(T, take(nextprimes(start), n))
 
 struct PrevPrimes{T<:Integer}
     start::T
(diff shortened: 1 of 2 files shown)


mschauer:
Is it justified to have a specialized function "iterate"? The canonical way to do this would be to "collect(take(...))".

rfourquet:
> Is it justified to have a specialized function "iterate"? The canonical way to do this would be to "collect(take(...))".

You are right, it's mostly out of laziness from me, as I didn't know how to portably use functions from `Base.Iterators` without warning, and thought it would be slower anyway. But it is not so slower (mostly negligible compared to testing `isprime`) and Compat can handle that after all. Updated accordingly.

src/Primes.jl
```diff
@@ -649,4 +655,102 @@ julia> prime(3)
...
+
+```
+julia> prevprimes(10, 3)
+3-element Array{Int64,1}:
+ 7
+ 5
+ 3
+```
+"""
+prevprimes{T<:Integer}(start::T, n::Integer) = collect(T, take(prevprimes(start, n)))
```
mschauer: ``take(prevprimes(start),n)`` ?

Keno:
I was just looking for something like this. The PR is a bit outdated, but seems like useful functionality?

rfourquet:
Good idea, I always postponed updating this branch!

src/Primes.jl
rfourquet: I'm not sure on this one. If you want an iterator, you might as well call `take` on `prevprime`, whereas at least this current version does *2* things! (`take` and `collect`). But we could perhaps just remove this function.

src/Primes.jl
```diff
@@ -528,6 +529,11 @@ function totient(n::Integer)
     totient(factor(abs(n)))
 end
 
+# add: checked add (when makes sense), result of same type as first argument
+
+add(n::BigInt, x::Int) = n+x
```
mschauer:
```suggestion
add(n::BigInt, x::Int) = n + x
```

src/Primes.jl
```diff
@@ -664,4 +670,116 @@ julia> prime(3)
...
+```
+"""
+nextprimes(start::T, n::Integer) where {T<:Integer} =
+    collect(T, Iterators.take(nextprimes(start), n))
+
+struct PrevPrimes{T<:Integer}
+    start::T
+end
+
+iterate(np::PrevPrimes, state=np.start) =
```
mschauer:
```suggestion
function iterate(np::PrevPrimes, state=np.start)
```

src/Primes.jl
```diff
@@ -664,4 +670,116 @@ julia> prime(3)
...
+    if isone(state)
+        nothing
+    else
+        p = prevprime(state)
+        (p, p-one(p))
+    end
```
mschauer:
```suggestion
    end
end
```

test/runtests.jl
```diff
@@ -390,3 +390,41 @@ for T in (Int, UInt, BigInt)
...
+
+@testset "prevprimes(::$T)" for T = (Int32, Int64, BigInt)
+    for (i, p) in enumerate(prevprimes(T(500)))
+        @test prevprime(T(500), i) == p
+        i > 20 && break
+    end
+    @test prevprimes(T(500), 10) == [prevprime(T(500), i) for i=1:10]
+    @test prevprimes(6, 1)[1] == prevprimes(5, 1)[1] == 5
+    @test prevprimes(4, 1)[1] == prevprimes(3, 1)[1] == 3
+    @test prevprimes(2, 1)[1] == 2
```
mschauer:
```suggestion
    @test prevprimes(2, 1)[1] == 2
    @test isempty(prevprimes(1, 1))
```

mschauer:
Some small suggestions.

src/Primes.jl
```diff
@@ -664,4 +670,116 @@ julia> prime(3)
...
+function iterate(np::NextPrimes, state=np.start)
+    p = nextprime(state)
+    (p, add(p, 1))
+end
+
+IteratorSize(::Type{<:NextPrimes}) = Base.IsInfinite()
+IteratorEltype(::Type{<:NextPrimes}) = Base.HasEltype()
+
+eltype(::Type{NextPrimes{T}}) where {T} = T
+
```
mschauer: Any idea why these 3 lines are not covered by the tests?

src/Primes.jl
```diff
@@ -664,4 +670,116 @@ julia> prime(3)
...
+    else
+        p = prevprime(state)
+        (p, p-one(p))
+    end
+
+IteratorSize(::Type{<:PrevPrimes}) = Base.SizeUnknown()
+IteratorEltype(::Type{<:PrevPrimes}) = Base.HasEltype()
+
+eltype(::Type{PrevPrimes{T}}) where {T} = T
+
```
mschauer: Same here.

src/Primes.jl
Keno: Probably a coverage bug.

Keno:
@rfourquet good to go?

rfourquet:
> good to go?

Let me review once more tomorrow and update with some comments from @mschauer.

src/Primes.jl
rfourquet: Oh I thought you were suggesting that it should return an iterator (i.e. remove the `collect`), when actually you pointed out that my parenthesis was incorrectly placed, right?

src/Primes.jl
rfourquet: I don't know, but I just added tests for `ItertatorEltype` and `IteratorSize`, maybe these will help decreasing less test coverage.

rfourquet:
This is good to go on my side.

PR #57, opened 2017-09-12
Fix depwarns on 0.7

yuyichao:


src/Primes.jl
```diff
@@ -178,7 +178,10 @@ julia> isprime(big(3))
 true
 ```
 """
-isprime(x::BigInt, reps=25) = ccall((:__gmpz_probab_prime_p,:libgmp), Cint, (Ptr{BigInt}, Cint), &x, reps) > 0
+isprime(x::BigInt, reps=25) = ccall((:__gmpz_probab_prime_p,:libgmp),
+                                    Cint, (Any, Cint), x, reps) > 0
+# TODO: Change `Any` to `Ref{BigInt}` when 0.6 support is dropped.
+# The two have the same effect but `Ref{BigInt}` won't be optimized on 0.6.
```
ararslan: `Any` is optimized but `Ref{BigInt}` isn't? That's kind of weird.

yuyichao: `Any` doesn't need any optimization for `!isbits`.

PR #61, opened 2018-06-29
upgradathon: require 0.7

rfourquet:


rfourquet:
There is a problem with a `mapreduce` deprecation on Julia master, I will update the code here when this is fixed on master.

README.md
```diff
@@ -1,6 +1,6 @@
 # Primes.jl
 
-[![Primes](http://pkg.julialang.org/badges/Primes_0.6.svg)](http://pkg.julialang.org/?pkg=Primes)
+[![Primes](http://pkg.julialang.org/badges/Primes_0.6.svg)](http://pkg.julialang.org/?pkg=Primes&ver=0.3)
```
ararslan: Was this intentional?

rfourquet: This is what I found at https://raw.githubusercontent.com/JuliaIO/JSON.jl/master/README.md, I _guessed_ it allows to test older versions of a package, but I mixed the versions and actually doesn't seem to work like that. What I wanted was to test with Julia version 0.6 the 0.3 release of Primes.jl. Do you have an idea? Should I update to `&ver=0.6` for Julia 0.6, or simply delete this line?

simonbyrne: I think that's for Julia versions (but pkg.julialang.org is only testing 0.6 at the moment, so it's hard to tell)

rfourquet: So shall I then just delete this line?

simonbyrne: Yes.

rfourquet: Ok thanks, done!

rfourquet:
I just updated with a more idiomatic way to delegate iteration (2 lines instead of 1).

src/Primes.jl
```diff
@@ -344,8 +338,13 @@ Set([2,5])
...
 factor(::Type{A}, n::T) where {T<:Integer, A<:AbstractArray} = A(factor(Vector{T}, n))
-factor(::Type{Vector{T}}, n::T) where {T<:Integer} =
-    mapreduce(collect, vcat, Vector{T}(), [repeated(k, v) for (k, v) in factor(n)])
+if VERSION >= v"0.7.0-beta.81"
+    factor(::Type{Vector{T}}, n::T) where {T<:Integer} =
+        mapreduce(collect, vcat, [repeated(k, v) for (k, v) in factor(n)], init=Vector{T}())
+else
+    factor(::Type{Vector{T}}, n::T) where {T<:Integer} =
+        mapreduce(collect, vcat, Vector{T}(), [repeated(k, v) for (k, v) in factor(n)])
+end
```
rfourquet: Would it be fine to keep only the new version, and require this specific julia version which implements it in REQUIRE? Would Appveyor and Travis be fine with that?

ararslan: If we were to do that, we would have to drop 0.7 from the testing matrix and just keep nightly, since 0.7 is 0.7.0-beta.0. I think it's fine to keep the conditional for now and drop it later.

rfourquet: Ok. I just pushed another fix for the just deprecated `ndigits`. Will merge later today if green.
Is it a good time afterwards to release version 0.4? may I do it?!

src/Primes.jl
```diff
@@ -1,20 +1,14 @@
 # This includes parts that were formerly a part of Julia. License is MIT: http://julialang.org/license
-__precompile__()
+#__precompile__()
```
ararslan: Why was precompilation disabled??

ararslan: Never mind, saw your follow-up commit.

rfourquet: Yes sorry for that; it's like re-reading 10 times an email, pressing the send button and spoting the mistake right away :persevere: (long life to "cancel send" buttons !)

ararslan: No worries, happens to us all!

PR #63, opened 2018-08-12
use * instead of ⋅ to show factorizations (fix #62)

### LINKED ISSUE #62: result of `factor` is no more copy-pastable
E.g.
```julia
julia> factor(20)
2^2 ⋅ 5
```
Before, the `⋅` operator (`\cdot`) was defined in `Base` (equivalent to `dot`), so, the expression `2^2 ⋅ 5` could be pasted back in the terminal and produce the input number, ie. `20`. Now it works only if `LinearAlgebra` is imported. I see few solutions:
1. document that `LinearAlgebra` must be imported for that to work
2. change `⋅` to `*` 
3. change `⋅` to `*` in the `show` method _when_ we can detect that `LinearAlgebra` is not loaded
4. have the `show(factor(20))` use `*` and keep `⋅` for the display at the REPL (suggested by @HarrisonGrodin, if I understood correctly)

I favor 1. and then 2.. I don't like 3 because the output depends on global state, and I find 4. to be slighly overkill. But after all, the star `*` is not so ugly, so 2. may be the most pragmatic solution. Other ideas?

rfourquet:


ararslan:
AppVeyor failure appears to be some network issue I guess?

PR #69, opened 2019-08-21
Primes.totient was not showing up in the documentation

reinermartin:
Thus I edited `api.md` to include `totient`. Also, I created a new subsection `Number-theoretic functions` (see  https://en.wikipedia.org/wiki/Arithmetic_function for terminology), where I put `totient` and `radical` under. I plan to add a whole bunch of such functions (which are easily computed from the prime decomposition of the given integer).


rfourquet:
Thanks!

PR #70, opened 2019-08-23
Adding four arithmetic functions

reinermartin:
I have added 4 functions: `moebius`, `liouville`, `divisorcount`, and `divisorsum`. They are all multiplicative  in the sense that ``f(n m) = f(n) f(m)`` if `n` and `m` are relative prime, and they can easily be computed from the prime factorization (see https://en.wikipedia.org/wiki/Arithmetic_function#Multiplicative_functions). Documentation and tests are included.

This is functionality for which a system like Mathematica has been useful in the past, but it would be good to have it in Julia. I plan to add more such functions and more detailed documentation later.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -33,5 +33,8 @@ Primes.primesmask
 ```@docs
 Primes.radical
 Primes.totient
-Primes.moebiusmu
+Primes.moebius
+Primes.liouville
+Primes.divisorcount
+Primes.divisorsum
 ```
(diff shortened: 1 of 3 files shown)


reinermartin:
I don't know how (technically) I can resolve this trivial conflict in api.md (the version where the 4 new functions are present is the valid one). Am I even allowed to do it? Help!

fkastner:
I'd also like to have this (esp. the Möbius `μ`). Any chance to revive this PR?

PR #71, opened 2019-08-25
Add an option to specify the increment to use when searching for primes

Keno:
A common operation in cryptography is to want to find a prime `p`
such that some `n` divides `p-1`. This is accomplished by starting
at some value `q` which `n` divides and then proceeding in increments
of `n` rather than increments of `1` until a prime is found. To enable
this procedure, add a keyword argument to `nextprime`/`prevprime` that
lets the user specify the desired increment.

Also, while we're at it, speed up the default `prevprime` case for larger
values by proceeding in increments of `2`.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -606,7 +606,7 @@ end
 
 
 """
-    prevprime(n::Integer, i::Integer=1; interval = 1)
+    prevprime(n::Integer, i::Integer=1; interval::Integer=1)
 
 The `i`-th largest prime not greater than `n` (in particular
 `prevprime(p) == p` if `p` is prime). If `i < 0`, this is equivalent to
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -544,13 +544,17 @@ end
 add_!(n::Integer, x::Int) = Base.checked_add(n, oftype(n, x))
 
 """
-    nextprime(n::Integer, i::Integer=1)
+    nextprime(n::Integer, i::Integer=1; interval = 1)
 
 The `i`-th smallest prime not less than `n` (in particular,
 `nextprime(p) == p` if `p` is prime). If `i < 0`, this is equivalent to
 prevprime(n, -i). Note that for `n::BigInt`, the returned number is
 only a pseudo-prime (the function [`isprime`](@ref) is used
 internally). See also [`prevprime`](@ref).
+
+If `interval` is provided, primes are sought in increments of `interval`.
+This can be useful to ensure the presince of certain divisors in `p-1`.
+The selected interval should be even.
 
 ```jldoctest
 julia> nextprime(4)
@@ -564,12 +568,19 @@ julia> nextprime(4, 2)
 
 julia> nextprime(5, 2)
 7
+
+julia> nextprime(2^16-1024+1; interval=1024)
+133121
+
+julia> gcd(133121 - 1, 1024) # 1024 | p - 1
+1024
 ```
 """
-function nextprime(n::Integer, i::Integer=1)
-    i < 0 && return prevprime(n, -i)
+function nextprime(n::Integer, i::Integer=1; interval = 1)
+    i < 0 && return prevprime(n, -i; interval=interval)
     i == 0 && throw(DomainError(i))
     n < 2 && (n = oftype(n, 2))
+    interval == 1 && (interval = 2)
     if n == 2
         if i <= 1
             return n
@@ -578,24 +589,24 @@ function nextprime(n::Integer, i::Integer=1)
             i -= 1
         end
     else
-        n += iseven(n)
+        n += iseven(n) ? oftype(n, interval - 1) : zero(n)
     end
     # n can now be safely mutated
     # @assert isodd(n) && n >= 3
     while true
         while !isprime(n)
-            n = add_!(n, 2)
+            n = add_!(n, interval)
         end
         i -= 1
         i <= 0 && break
-        n = add_!(n, 2)
+        n = add_!(n, interval)
     end
     n
 end
 
 
 """
-    prevprime(n::Integer, i::Integer=1)
+    prevprime(n::Integer, i::Integer=1; interval = 1)
 
 The `i`-th largest prime not greater than `n` (in particular
 `prevprime(p) == p` if `p` is prime). If `i < 0`, this is equivalent to
@@ -614,17 +625,24 @@ julia> prevprime(5, 2)
 3
 ```
 """
-function prevprime(n::Integer, i::Integer=1)
-    i <= 0 && return nextprime(n, -i)
-    n += zero(n) # deep copy of n, which is mutated below
+function prevprime(n::Integer, i::Integer=1; interval = 1)
+    i <= 0 && return nextprime(n, -i; interval=interval)
+    i == 1 && n == 2 && return n
+    # A bit ugly, but this lets us speed up prime walking 2x in the (common)
+    # case that n >> 2, while preventing prevprime(3) from skipping 2 and giving
+    # the wrong answer.
+    was_one = interval == 1
+    was_one && (interval = 2)
+    n -= iseven(n) ? oftype(n, interval-1) : zero(n) # deep copy of n, which is mutated below
     while true
         n < 2 && throw(ArgumentError("There is no prime less than or equal to $n"))
+        was_one && n <= 4 && (interval = 1)
         while !isprime(n)
-            n = add_!(n, -1)
+            n = add_!(n, -interval)
         end
         i -= 1
         i <= 0 && break
-        n = add_!(n, -1)
+        n = add_!(n, -interval)
     end
     n
 end
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -544,7 +544,7 @@ end
 add_!(n::Integer, x::Int) = Base.checked_add(n, oftype(n, x))
 
 """
-    nextprime(n::Integer, i::Integer=1; interval = 1)
+    nextprime(n::Integer, i::Integer=1; interval::Integer=1)
 
 The `i`-th smallest prime not less than `n` (in particular,
 `nextprime(p) == p` if `p` is prime). If `i < 0`, this is equivalent to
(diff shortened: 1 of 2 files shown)


src/Primes.jl
```diff
@@ -544,14 +544,18 @@ end
 add_!(n::Integer, x::Int) = Base.checked_add(n, oftype(n, x))
 
 """
-    nextprime(n::Integer, i::Integer=1)
+    nextprime(n::Integer, i::Integer=1; interval = 1)
```
ararslan:
```suggestion
    nextprime(n::Integer, i::Integer=1; interval::Integer=1)
```

src/Primes.jl
```diff
@@ -544,14 +544,18 @@ end
...
 
 The `i`-th smallest prime not less than `n` (in particular,
 `nextprime(p) == p` if `p` is prime). If `i < 0`, this is equivalent to
 prevprime(n, -i). Note that for `n::BigInt`, the returned number is
 only a pseudo-prime (the function [`isprime`](@ref) is used
 internally). See also [`prevprime`](@ref).
 
+If `interval` is provided, primes are sought in increments of `interval`.
+This can be useful to ensure the presince of certain divisors in `p-1`.
```
ararslan:
```suggestion
This can be useful to ensure the presence of certain divisors in `p-1`.
```

src/Primes.jl
```diff
@@ -564,12 +568,19 @@ julia> nextprime(4, 2)
...
+julia> nextprime(2^16-1024+1; interval=1024)
+133121
+
+julia> gcd(133121 - 1, 1024) # 1024 | p - 1
+1024
 ```
 """
-function nextprime(n::Integer, i::Integer=1)
-    i < 0 && return prevprime(n, -i)
+function nextprime(n::Integer, i::Integer=1; interval = 1)
```
ararslan:
```suggestion
function nextprime(n::Integer, i::Integer=1; interval::Integer=1)
```

src/Primes.jl
```diff
@@ -578,24 +589,24 @@ function nextprime(n::Integer, i::Integer=1)
...
-        n = add_!(n, 2)
+        n = add_!(n, interval)
     end
     n
 end
 
 
 """
-    prevprime(n::Integer, i::Integer=1)
+    prevprime(n::Integer, i::Integer=1; interval = 1)
```
ararslan:
```suggestion
    prevprime(n::Integer, i::Integer=1; interval::Integer=1)
```

src/Primes.jl
```diff
@@ -614,17 +625,24 @@ julia> prevprime(5, 2)
 3
 ```
 """
-function prevprime(n::Integer, i::Integer=1)
-    i <= 0 && return nextprime(n, -i)
-    n += zero(n) # deep copy of n, which is mutated below
+function prevprime(n::Integer, i::Integer=1; interval = 1)
```
ararslan:
```suggestion
function prevprime(n::Integer, i::Integer=1; interval::Integer=1)
```

src/Primes.jl
```diff
@@ -564,12 +568,19 @@ julia> nextprime(4, 2)
 
 julia> nextprime(5, 2)
 7
+
+julia> nextprime(2^16-1024+1; interval=1024)
+133121
```
rfourquet: This doesn't seem correct... For once, `isprime(2^16-1024+1) == true`, and then `nextprime(2^16-1024+1, 2; interval=1024) == 2^16+1` is prime.

src/Primes.jl
```diff
@@ -578,24 +589,24 @@ function nextprime(n::Integer, i::Integer=1)
             i -= 1
         end
     else
-        n += iseven(n)
+        n += iseven(n) ? oftype(n, interval - 1) : zero(n)
```
rfourquet: This is incorrect: `n` should only be incremented my (multiples of) `interval`. I guess you wanted to prevent e.g. `nextprime(10, interval=4)` turning into an infinite loop, but instead a check for two even values for `n`  and `interval` could be made upfront and throw.

src/Primes.jl
```diff
@@ -564,12 +568,19 @@ julia> nextprime(4, 2)
...
+1024
 ```
 """
-function nextprime(n::Integer, i::Integer=1)
-    i < 0 && return prevprime(n, -i)
+function nextprime(n::Integer, i::Integer=1; interval::Integer=1)
+    i < 0 && return prevprime(n, -i; interval=interval)
     i == 0 && throw(DomainError(i))
     n < 2 && (n = oftype(n, 2))
+    interval == 1 && (interval = 2)
```
rfourquet: Why not having directly `interval=2` by default?

src/Primes.jl
```diff
@@ -544,14 +544,18 @@ end
...
 
 The `i`-th smallest prime not less than `n` (in particular,
 `nextprime(p) == p` if `p` is prime). If `i < 0`, this is equivalent to
 prevprime(n, -i). Note that for `n::BigInt`, the returned number is
 only a pseudo-prime (the function [`isprime`](@ref) is used
 internally). See also [`prevprime`](@ref).
 
+If `interval` is provided, primes are sought in increments of `interval`.
+This can be useful to ensure the presence of certain divisors in `p-1`.
+The selected interval should be even.
```
rfourquet: This should be clarified as to why? And maybe put a test in the function? But I believe this restriction is unnecessary...

src/Primes.jl
```diff
@@ -578,24 +589,24 @@ function nextprime(n::Integer, i::Integer=1)
             i -= 1
```
rfourquet: Github doesn't allow me to comment one line higher: `n` should be incremented by `interval` (provided we allow an `isodd` interval).

src/Primes.jl
```diff
@@ -578,24 +589,24 @@ function nextprime(n::Integer, i::Integer=1)
...
-        n = add_!(n, 2)
+        n = add_!(n, interval)
     end
     n
 end
 
 
 """
-    prevprime(n::Integer, i::Integer=1)
+    prevprime(n::Integer, i::Integer=1; interval::Integer=1)
```
rfourquet: `interval` should be documented, or at least say to lookup `nextprime` for explanation.

src/Primes.jl
```diff
@@ -614,17 +625,24 @@ julia> prevprime(5, 2)
...
-    n += zero(n) # deep copy of n, which is mutated below
+function prevprime(n::Integer, i::Integer=1; interval::Integer=1)
+    i <= 0 && return nextprime(n, -i; interval=interval)
+    i == 1 && n == 2 && return n
+    # A bit ugly, but this lets us speed up prime walking 2x in the (common)
+    # case that n >> 2, while preventing prevprime(3) from skipping 2 and giving
+    # the wrong answer.
+    was_one = interval == 1
+    was_one && (interval = 2)
+    n -= iseven(n) ? oftype(n, interval-1) : zero(n) # deep copy of n, which is mutated below
```
rfourquet: Similarly, should decrement by `interval`. For example, this makes `prevprime(10, interval=4) == 7` instead of the correct answer `2`.

src/Primes.jl
```diff
@@ -614,17 +625,24 @@ julia> prevprime(5, 2)
 3
 ```
 """
-function prevprime(n::Integer, i::Integer=1)
-    i <= 0 && return nextprime(n, -i)
-    n += zero(n) # deep copy of n, which is mutated below
+function prevprime(n::Integer, i::Integer=1; interval::Integer=1)
+    i <= 0 && return nextprime(n, -i; interval=interval)
+    i == 1 && n == 2 && return n
+    # A bit ugly, but this lets us speed up prime walking 2x in the (common)
```
rfourquet: Good catch for this pre-existing unefficiency!

src/Primes.jl
```diff
@@ -544,14 +544,18 @@ end
 add_!(n::Integer, x::Int) = Base.checked_add(n, oftype(n, x))
 
 """
-    nextprime(n::Integer, i::Integer=1)
+    nextprime(n::Integer, i::Integer=1; interval::Integer=1)
```
rfourquet: The name of your git branch suggests that `interval` could be named `increment`, which sounds better... but the problem is that for `prevprime` it would have to be called `decrement`. Thoughts?

src/Primes.jl
```diff
@@ -564,12 +568,19 @@ julia> nextprime(4, 2)
...
+julia> gcd(133121 - 1, 1024) # 1024 | p - 1
+1024
 ```
 """
-function nextprime(n::Integer, i::Integer=1)
-    i < 0 && return prevprime(n, -i)
+function nextprime(n::Integer, i::Integer=1; interval::Integer=1)
+    i < 0 && return prevprime(n, -i; interval=interval)
     i == 0 && throw(DomainError(i))
     n < 2 && (n = oftype(n, 2))
```
rfourquet: `n` should only be incremented by multiples of `interval`.

src/Primes.jl
```diff
@@ -614,17 +625,24 @@ julia> prevprime(5, 2)
...
+    # the wrong answer.
+    was_one = interval == 1
+    was_one && (interval = 2)
+    n -= iseven(n) ? oftype(n, interval-1) : zero(n) # deep copy of n, which is mutated below
     while true
         n < 2 && throw(ArgumentError("There is no prime less than or equal to $n"))
+        was_one && n <= 4 && (interval = 1)
         while !isprime(n)
-            n = add_!(n, -1)
+            n = add_!(n, -interval)
```
rfourquet: `add_!` is implemented when the 2nd arg is an `Int`, so more methods must be written, or `interval` restricted to `Int` (applies also to `nextprime`).

rfourquet:
@ararslan Please, give us other reviewers a bit more time to review before merging, and even more on the WE! 
I would wish at least two days for non trivial changes.

src/Primes.jl
rfourquet: Agreed :)

rfourquet:
> Sorry, will do next time.

Thanks, and no worries :)

PR #74, opened 2019-09-17
factorization setindex! types cleanup

milesfrain:
Are there any reasons for how types are used in the original version?

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -27,12 +27,12 @@ end
 
 Base.getindex(f::Factorization, p::Integer) = get(f, p, 0)
 
-function Base.setindex!(f::Factorization{T}, e::Int, p::Integer) where T
+function Base.setindex!(f::Factorization{T}, e::Int, p::T) where T<:Integer
     found = searchsorted(f.pe, p, by=first)
     if isempty(found)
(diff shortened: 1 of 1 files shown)


src/factorization.jl
```diff
@@ -27,12 +27,12 @@ end
 
 Base.getindex(f::Factorization, p::Integer) = get(f, p, 0)
 
-function Base.setindex!(f::Factorization{T}, e::Int, p::Integer) where T
+function Base.setindex!(f::Factorization{T}, e::Int, p::T) where T<:Integer
```
ararslan: This restricts the values that are accepted by the method to only those where the type of `p` exactly matches the element type of the factorization. Currently we allow any kind of integer, and conversion when inserting into the factorization object happens automatically. Did you find a case where that doesn't work as expected?

milesfrain: Thank for the clarification. I assumed it would be better to enforce that the value matches the factorization type, rather than rely on a conversion. Hopefully this is compile-time conversion.

I should have tested this first with something like `Primes.factor(UInt(36))` to see how this change breaks a lot of stuff that's hardcoded to use `Int`.

`UInt` seems like a more appropriate datatype to use for factoring and primes, but I guess that would be more annoying for general usages.

PR #75, opened 2019-11-18
Fix prevprime for unsigned integers

Keno:


src/Primes.jl
```diff
@@ -533,6 +533,8 @@ end
 
 add(n::BigInt, x::Int) = n + x
 add(n::Integer, x::Int) = Base.checked_add(n, oftype(n, x))
+sub(n::BigInt, x::Int) = n - x
```
ararslan: Coverage drop is apparently because this line isn't hit by the tests.

Keno: True, the line is not used, because it gets intercepted below. I like having it for symmetry though.

PR #77, opened 2019-11-22
allow `primes(..., typemax(Int))` without throwing an error

### LINKED ISSUE #76: `primes` throws exception, when arguments are huge.
The second call doesn't look fine. The arguments are close to `typemax(Int)`.

```
julia> primes(2^62-100, 2^62-1)
2-element Array{Int64,1}:
 4611686018427387817
 4611686018427387847

julia> primes(2^63-100, 2^63-1)
ERROR: BoundsError: attempt to access 31-element Array{Int64,1} at index [-26]
Stacktrace:
 [1] getindex at ./array.jl:758 [inlined]
 [2] wheel_index at /home/crusius/.julia/packages/Primes/uaYlp/src/Primes.jl:26 [inlined]
 [3] _primesmask(::Int64, ::Int64) at /home/crusius/.julia/packages/Primes/uaYlp/src/Primes.jl:68
 [4] primes(::Int64, ::Int64) at /home/crusius/.julia/packages/Primes/uaYlp/src/Primes.jl:119
 [5] top-level scope at REPL[91]:1
```

KlausC:
Fixes #76.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -63,9 +63,11 @@ function _primesmask(lo::Int, hi::Int)
         if small_sieve[i]
             p = wheel_prime(i)
             j = wheel_index(2 * div(lo - p - 1, 2p) + 1)
-            q = p * wheel_prime(j + 1)
+            r = widemul(p, wheel_prime(j + 1))
+            r > m && continue
             j = j & 7 + 1
-            while q ≤ m
+            q = Int(r)
+            while 0 ≤ q ≤ m
                 sieve[wheel_index(q) - wlo] = false
                 q += wheel[j] * p
                 j = j & 7 + 1
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -64,9 +64,10 @@ function _primesmask(lo::Int, hi::Int)
             p = wheel_prime(i)
             j = wheel_index(2 * div(lo - p - 1, 2p) + 1)
             r = widemul(p, wheel_prime(j + 1))
-            r > m && continue
+            r > m && continue # use widemul to avoid r <= m caused by overflow
             j = j & 7 + 1
             q = Int(r)
+            # q < 0 indicates overflow when incrementing q inside loop
             while 0 ≤ q ≤ m
                 sieve[wheel_index(q) - wlo] = false
                 q += wheel[j] * p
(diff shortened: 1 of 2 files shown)


KlausC:
gentle reminder...

KlausC:
bump @ararslan

rfourquet:
Sorry this took so long. I will try to review by the end of next week, otherwise don't hesitate to ping me.

KlausC:
@rfourquet  wanted to be pinged

src/Primes.jl
```diff
@@ -63,9 +63,11 @@ function _primesmask(lo::Int, hi::Int)
         if small_sieve[i]
             p = wheel_prime(i)
             j = wheel_index(2 * div(lo - p - 1, 2p) + 1)
-            q = p * wheel_prime(j + 1)
+            r = widemul(p, wheel_prime(j + 1))
```
rfourquet, line 66: Instead of computing with `widemul`, wouldn't it be the case that keeping `q = p*wheel_prime(j+1)` and checking `q < 0 && continue` would be enough? Anyway, I'm just curious, it's clearer and probably better (i.e. not a performance proablem) to use `widemul` as arguments are only `Int` so `BigInt` are not involved in `widemul`.

src/Primes.jl
```diff
@@ -63,9 +63,11 @@ function _primesmask(lo::Int, hi::Int)
...
+            r > m && continue
             j = j & 7 + 1
-            while q ≤ m
+            q = Int(r)
+            while 0 ≤ q ≤ m
```
rfourquet, line 70: Maybe add a comment that the `0 <= q` part is an overflow check (as it's less obvious here than above where `widemul` is used).

rfourquet:
Thanks for you PR! I finally got time to refresh my memory on this "wheel" thing. I just think that a comment could be added but this LGTM.

src/Primes.jl
KlausC: That is right, when adding relatively small integers to `q`, overflow is indicated by `q < 0`. I will make a comment to clarify.

src/Primes.jl
KlausC: I made quite some experiments with that, but found `q < 0` was not a reliable indicator for overflow of `*`; that is why the `widemul`. I am not sure, why I did not use an overflow-checked variant of `*`, probably not more efficient.

rfourquet: Yes `Base.mul_with_overflow` comes to mind, but `widemul` seems to be plenty fast enough in this case, and I didn't notice performance regressions.

PR #78, opened 2019-11-29
remove REQUIRE, add [compat], upgrade version

rfourquet:
Tagging a new version seems long overdue. I added a compat section, and tested manually `DataStructures` to find the lowest supported version, which is `v"0.11.0`. Which I put in the `[compat]` section, all the way up to the current `v"1.17"`. But it looks suboptimal to have to list all the individual intermediate versions, is there a shortcut? (I didn't find documentation on using inequalities with a `&&`).


Project.toml
```diff
@@ -1,9 +1,14 @@
 name = "Primes"
 uuid = "27ebfcd6-29c5-5fa9-bf4b-fb8fc14df3ae"
+version = "0.5.0"
```
ararslan: The package is stable enough that at this point I'd argue we could just declare the version to be 1.0.0.

rfourquet: I had a thought along the same lines, and ended up choosing the easier path of just staying beta, and postpone 1.0 for next time! Let's see if someone else has an opinion. Maybe @mschauer ?

mschauer: We can make it a 1.0 👍

Project.toml
```diff
@@ -1,9 +1,14 @@
...
 [extras]
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
 test = ["DataStructures", "Test"]
+
+[compat]
+DataStructures = "0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17"
+julia = "1.0.0"
```
ararslan:
```suggestion
julia = "1"
```

ararslan:
> But it looks suboptimal to have to list all the individual intermediate versions, is there a shortcut?

Unfortunately no.

Project.toml
rfourquet: Ok sounds good.

I have just one hesitation sorry, it's about our `factor(ContainerType, n::Integer)` API. It doesn't seem very "Julian", for example for a dict type `D` we have to do `factor!(n, D(Dict{T,Int}()))`,  because we don't know in general how to create an empty instance of `D` (this one isn't even guaranteed to succeed). I'm probably the one who introduced it in the first place, but I feel it would be nicer to switch to something like `factor!(d::AbstractDict, n::Integer)` (`factor!` is not exported, so we could switch the order of the arguments, to respect the guidelines). On the other hand, it's not a big deal to continue supporting indefinitely `factor(ContainerType, n::Integer)` if we tag `v"1.0.0"`. What do you think?

mschauer: Absolutely, I share that view. The best is the enemy of the good, but if we actually want to fix it, why not now?

rfourquet: So I did a bit of history digging, initially we had `factorvec(n)` in #13 by @ararslan , which I modified to `factor(Vector, n)` in #19, because I apparently didn't like the name `vectorvec` :D 
If we now move to `factor!(Vector{Int}(), n)` or `factor!(Int[], n)`: we are forced to specify the type of the array if we want a tight-typed array. In any case, this will remove a bit of the convenience of the initial `factorvec`. I think this is OK for me, and we can still revisit this issue in the future (this issue reminds a feature request by Chris Rackauckas in the Julia repo related to specifying how to instantiate some arrays from a set of parameters, but I might be remembering incorrectly). 

A possible future way would be: `factor!([n])`; I have a reasoning for that, but this belong to its own PR.

In any case, I'm pretty sure I want the `factor!(containerinstance, n::Integer)` API (where `containerinstance` is allowed to already contain something). So I will open a PR with it, to replace the current `ContainerType` one (which could be deprecated, or maybe this is overkill, because precisely we were in pre-v1).

I could have put this comment in the new PR directly, but I needed this thinking aloud to come to this conclusion ;-)

rfourquet:
I will go ahead and merge this one. I still think we should have another round of discussion about the possibility to deprecate one of our API before tagging Primes v1.0, but we shouldn't hold-up other changes which should be released. I will try to be more pro-active in the near future about getting to v1.0.

PR #79, opened 2019-11-29
CI: test on julia 1.2, 1.3, and allow nightly failure

rfourquet:
I'm not sure if the failure in #78 is my responsibility, if it isn't I suggest allowing nightly to fail in CI. In any case, we should now test 1.3 and probably 1.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ julia:
   - 1.1
   - 1.2
   - 1.3
+  - 1.4
   - nightly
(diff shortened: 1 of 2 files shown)


rfourquet:
This same failure here indicates that this is indeed probably a bug in `Pkg`, which Kristoffer also [suggests](https://github.com/JuliaLang/Pkg.jl/pull/1444#issuecomment-559705686).

PR #84, opened 2020-06-15
primes: accept non-Int integer types as argument(s)

rfourquet:
`primes` seems to be our only function currently accepting only `Int`.


rfourquet:
I will merge tomorrow if no objection by then.

ghislain-bernard:
Hi, any news about this request ?

PR #85, opened 2020-06-15
remove broken badge from the README

rfourquet:


rfourquet:
@ararslan it seems like you initially put this badge, I don't remember what its purpose is, but I will let you add an alternative if you see one.

ararslan:
Oh, that was an old PackageEvaluator badge. Okay to remove.

PR #86, opened 2020-06-19
update Primes.jl

deva567:
Earlier primes(0) will through error like "The condition lo ≤ hi must be met." but now
primes(0) will return output as look like primes(0,0).

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -112,6 +112,9 @@ primesmask(n::Integer) = n ≤ typemax(Int) ? primesmask(Int(n)) :
 Returns a collection of the prime numbers (from `lo`, if specified) up to `hi`.
 """
 function primes(lo::Int, hi::Int)
+    if hi<1
+    list=Int[]
+    else
     lo ≤ hi || throw(ArgumentError("The condition lo ≤ hi must be met."))
     list = Int[]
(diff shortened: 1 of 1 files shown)


rfourquet:
Thanks! This would need a bit more work to be merged: for changes like this, we need tests to avoid regressions. Also, given that e.g. `primes(-10, -4)` works, I believe we should update `primes(n)` such that this never throws, i.e. even for `n<0`, not only `n==0`.

deva567:
@rfourquet  just for clarification , we need to produce output for primes(-10,-4) as same as like primes(4,10) and one more thing primes(-10) whether it should equal to primes(0,10).
FYI attached sample PNG files 

sample1

sample

rfourquet:
No, we have this method: `primes(n) = primes(1, n)`, but conceptually this would be `primes(n) = primes(-infinity, n)`. So for `m <= n <1`,  we have `isempty(primes(m, n)) == true`, i.e. no primes are negative. So `primes(n)` for `n < 1` should just return the empty list as well.

deva567:
@rfourquet  
one1
FYI attached sample output  and made changes in Primes.jl with the help of **"if -else"**  now it can handle negative integer when we gave one parameter to the function and wont through error like earlier

PR #87, opened 2020-09-07
Segmented prime sieve for fun and profit

### LINKED ISSUE #45: Reduce memory usage in the sieve
Currently, memory usage of the prime sieve increases linearly with the size of the interval `[lo, hi]`, while the number of primes only grows asymptotically as `(hi) / log (hi)`. Furthermore, L1 memory caching is not exploited.

By implementing a segmented sieve both problems can be addressed at once: lowering memory consumption and increasing speed (in a C++ example I wrote up to 10x).

The current algorithm is basically:

1. Discover prime numbers `p` less than `√hi`
2. Cross off multiples of `p` starting at `p * p`.

The suggested procedure is:

1. Discover all prime numbers `p` less than `√hi`
2. Divide [√hi, hi] into fixed size segments / smaller intervals (fitting L1 cache)
3. Outer loop: iterate over each segment
   - Inner loop: cross off all multiples of all primes in the current segment
   - Collect all primes in the current segment

You only need to store one segment (constant memory allocation) in step 2/3 and if the segment fits in L1 cache it is extremely fast, since you're iterating over it multiple times when sieving.

For reference see http://primesieve.org/

haampie:
Closes https://github.com/JuliaMath/Primes.jl/issues/45.

I have this code lying around for a bit now, but never turned it into a PR. Now I'm turning it into a PR to force myself to finish it :sweat_smile:. It could use some tests and maybe some code style improvements / cleanup.

What it brings is an iterator:

```julia
using Primes.SegmentedSieve: SegmentIterator, vec_count_ones

function count_primes(a, b)
    cnt = 0
    for segment in SegmentIterator(a:b, 32 * 1024)
        cnt += vec_count_ones(segment.segment)
    end
    return cnt
end
```

which makes some operations very fast, as they operate on O(1) memory instead of O(n). For reference:

```julia
julia> @time count(Primes.primesmask(2^31, 2^32))
  7.362490 seconds (3.86 k allocations: 802.346 MiB, 1.18% gc time)
98182656

julia> @time count_primes(2^31, 2^32)
  0.452503 seconds (2.19 k allocations: 375.266 KiB)
98182656
```

this is 16x faster.

There's some neat optimizations: you can pick a buffer size to make things fit in L1 cache (above it's 32kb); the inner loop is unrolled by hand such that 8 multiples are removed per iteration; memory is compressed s.t. 1 byte equals an interval of 30 numbers (through a {2, 3, 5}-wheel); multiples of {7, 11, 13, 17} are "pre-sieved" in the sense that a buffer of size 7 * 11 * 13 * 17 is sieved and repeatedly copied over, instead of actually sieving 7, 11, 13 and 17 each over the full range.

You can also print the segment if you like:

```julia
julia> for segment in SegmentIterator(1:180, 2)
           println(segment)
       end
    01  07  11  13  17  19  23  29  
00   .   x   x   x   x   x   x   x  
30   x   x   x   x   x   .   x   x  

    01  07  11  13  17  19  23  29  
60   x   x   x   x   .   x   x   x  
90   .   x   x   x   x   x   x   .  

     01  07  11  13  17  19  23  29  
120   .   x   x   .   x   x   .   x  
150   x   x   .   x   x   .   x   x 
```
So, here it has a buffer of 2 bytes, resembling an interval of 60 numbers, meaning it needs 3 iterations. Every `x` marks a prime number.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -9,9 +9,11 @@ using Base: BitSigned
 using Base.Checked: checked_neg
 
 export isprime, primes, primesmask, factor, ismersenneprime, isrieselprime,
-       nextprime, nextprimes, prevprime, prevprimes, prime, prodfactors, radical, totient
+       nextprime, nextprimes, prevprime, prevprimes, prime, prodfactors, radical, totient,
+       SegmentedSieve
 
 include("factorization.jl")
+include("segmented_sieve/SegmentedSieve.jl")
 
 # Primes generating functions
 #     https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes
--- a/src/segmented_sieve/SegmentedSieve.jl
+++ b/src/segmented_sieve/SegmentedSieve.jl
@@ -1,0 +1,44 @@
+module SegmentedSieve
+
+const ps = (1, 7, 11, 13, 17, 19, 23, 29)
+
+"""
+Population count of a vector of UInt8s for counting prime numbers.
+See https://github.com/JuliaLang/julia/issues/34059
+"""
+function vec_count_ones(xs::Union{Vector{UInt8}, Base.FastContiguousSubArray{UInt8}})
+    n = length(xs)
+    count = 0
+    chunks = n ÷ sizeof(UInt)
+    GC.@preserve xs begin
+        ptr = Ptr{UInt}(pointer(xs))
+        for i in 1:chunks
+            count += count_ones(unsafe_load(ptr, i))
+        end
+    end
+
+    @inbounds for i in 8chunks+1:n
+        count += count_ones(xs[i])
+    end
+
+    count
+end
+
+function to_idx(x)
+    x ==  1 && return 1
+    x ==  7 && return 2
+    x == 11 && return 3
+    x == 13 && return 4
+    x == 17 && return 5
+    x == 19 && return 6
+    x == 23 && return 7
+    return 8
+end
+
+include("generate_sieving_loop.jl")
+include("sieve_small.jl")
+include("presieve.jl")
+include("siever.jl")
+include("sieve.jl")
+
+end # module
--- a/src/segmented_sieve/generate_sieving_loop.jl
+++ b/src/segmented_sieve/generate_sieving_loop.jl
@@ -1,0 +1,145 @@
+"""
+For a prime number p and a multiple q, the wheel index encodes the prime number index of 
+p and q modulo 30, which can be encoded to a single number from 1 ... 64. This allows us
+to jump immediately into the correct loop at the correct offset.
+"""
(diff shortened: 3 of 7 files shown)


galanakis:
Is there a documentation or a paper this algorithm is based on? I would like to help review it.

haampie:
I think primesieve.org is a good reference. The wheel is a generalization of the idea of skipping multiples of 2 (which is easy to code), see https://en.wikipedia.org/wiki/Wheel_factorization for a good overview.

If you take a {2, 3, 5}-wheel, then consider any number `z = 2*3*5*x + y = 30*x + y`, where y = 1, ..., 30. Then z is composite if y is a multiple of 2, 3 or 5. So you only have to verify y = {1, 7, 11, 13, 17, 19, 23, 29} are prime for all x. Those are 8 numbers, and them being prime or not is 1 bit of info, so this neatly compressed to 1 byte of data per interval of 30 numbers on the number line.

The next optimization after compressing the number line by a constant factor is to work on constant size segments of the number line. So instead of `1:n`, you work in segments `1:b`, `b+1:2b`, `2b+1:3b`, ..., `m*b+1:n`, and you allocate the boolean array once and for all. Effectively you do:

```julia
for (a, b) in segments # sieve a:b
  for p in primes_so_far
    # find the smallest number q st p * q >= a.
    # then remove p * q, p * (q + 1), p * (q + 2), ... until p * (q + m) > b.
  end
end
```

the inner loop is finicky, cause (a) you don't want to always compute the smallest q s.t. p * q >= a, instead you already *know* what it is thanks to the last iteration in the previous segment. and (b) you aren't going to remove multiples q, q+1, q+2, ... as you're skipping all p * (q+i)'s that aren't in the {1, 7, 11, 13, 17, 19, 23, 29} spokes -- and this is where the loop is unrolled by hand, and where all the pain and suffering comes from.

galanakis:
Thank you. I am already familiar with wheel factorization and the the Segmented Sieve (Bays-Hudson). I have to think a bit about combining the two. I appreciate the long response and I will think about it.

galanakis:
I have not thought about it very much, but this paper has some other little optimizations, under Algorithms B and C.

https://link.springer.com/content/pdf/10.1007/BF01932283.pdf

Is your segmented sieve similar to this? Or are you using a completely different approach which is incompatible. Maybe it is incompatible, because this paper basically uses the usual 2-wheel, but I really have to think about it.

oscardssmith:
Want to merge update and this? It's a lot of extra code, but I've become convinced that something like this is necessary for optimal performance.

oscardssmith:
Want to merge update and this? It's a lot of extra code, but I've become convinced that something like this is necessary for optimal performance.

PR #93, opened 2021-02-25
Remove `isprime` from trial division part of `factor`

oscardssmith:
The theory here is that when the `factor` is used on a large number with many small factors, it ends up calling `isprime` about 5000 times. This is incredibly wasteful. For numbers where the `isprime` is true, this will still detect it relatively quickly. There is a 99.7% chance that this change will be less than 4 `mod` operations slower. As such, this is probably very slightly slower for factoring a small number times a massive prime, but much faster for everything else.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -265,9 +265,14 @@ function factor!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Integer}
                 n = div(n, p)
             end
             n == 1 && return h
-            isprime(n) && (h[n] = 1; return h)
+            if T==BigInt 
+                isprime(n, reps=1) && (h[n] = 1; return h)
+            else
+                isprime(n) && (h[n] = 1; return h)
+            end
         end
     end
+    isprime(n) && h[n]=1; return h
     T <: BigInt || widemul(n - 1, n - 1) ≤ typemax(n) ? pollardfactors!(n, h) : pollardfactors!(widen(n), h)
 end
 
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -272,7 +272,7 @@ function factor!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Integer}
             end
         end
     end
-    isprime(n) && (h[n]=1; return h)
+    T==BigInt && isprime(n) && (h[n]=1; return h)
     T <: BigInt || widemul(n - 1, n - 1) ≤ typemax(n) ? pollardfactors!(n, h) : pollardfactors!(widen(n), h)
 end
 
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -266,7 +266,7 @@ function factor!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Integer}
             end
             n == 1 && return h
             if T==BigInt 
-                isprime(n, reps=1) && (h[n] = 1; return h)
+                isprime(n, 1) && (h[n] = 1; return h)
             else
                 isprime(n) && (h[n] = 1; return h)
             end
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -256,7 +256,9 @@ function factor!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Integer}
     end
 
     local p::T
+    nsqrt = isqrt(n)
     for p in PRIMES
+        p > nsqrt && break
         if n % p == 0
             h[p] = get(h, p, 0) + 1
             n = div(n, p)
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -267,6 +267,7 @@ function factor!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Integer}
                 n = div(n, p)
             end
             n == 1 && return h
+            nsqrt = isqrt(n)
         end
     end
     isprime(n) && (h[n]=1; return h)
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -266,7 +266,7 @@ function factor!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Integer}
             end
             n == 1 && return h
             if T==BigInt 
-                isprime(n, 1) && (h[n] = 1; return h)
+                isprime(n, 4) && (h[n] = 1; return h)
             else
                 isprime(n) && (h[n] = 1; return h)
             end
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -265,14 +265,9 @@ function factor!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Integer}
                 n = div(n, p)
             end
             n == 1 && return h
-            if T==BigInt 
-                isprime(n, 4) && (h[n] = 1; return h)
-            else
-                isprime(n) && (h[n] = 1; return h)
-            end
         end
     end
-    T==BigInt && isprime(n) && (h[n]=1; return h)
+    isprime(n) && (h[n]=1; return h)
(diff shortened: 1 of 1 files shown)


src/Primes.jl
```diff
@@ -265,9 +265,14 @@ function factor!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Integer}
                 n = div(n, p)
             end
             n == 1 && return h
-            isprime(n) && (h[n] = 1; return h)
+            if T==BigInt 
+                isprime(n, 1) && (h[n] = 1; return h)
```
rfourquet, line 269: How does it compare with using `reps==2` or `3` here? Naively, one imagines that that could be not much more expensive (compared to the initial `25`), while providing more correct answer. (IOW, the false positive rate is `0.25^reps`, how does the runtime of `isprime` scale with `reps` in the case of the intended improvement?)

src/Primes.jl
```diff
@@ -265,9 +265,14 @@ function factor!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Integer}
...
+            else
+                isprime(n) && (h[n] = 1; return h)
+            end
         end
     end
+    isprime(n) && (h[n]=1; return h)
```
rfourquet, line 275: Does it not need to happen only for `T == BigInt` ?

oscardssmith: `isprime` with reps only works for `BigInt`.

src/Primes.jl
oscardssmith: This probably should be higher. When I initially made the change, I thought this was checking `isprime` after every trial division, but it actually only happens after successful ones. As such, I think changing this to `4` probably makes sense.

src/Primes.jl
rfourquet: What I mean is that above, the code was changed only for `BigInt`, so it seems that the code should not be changed here for non-`BigInt`.

oscardssmith: Oh yeah. Will fix.

oscardssmith: Also, would it be hard to get a native julia version of `isprime(x,reps)`? That would be the cleaner fix by a bunch.

oscardssmith:
Anything need to be done to merge this?

oscardssmith:
Bumping. Can we get this merged?

rfourquet:
So this looks good in theory, but I didn't manage to find practical runtime improvements. E.g. `x =  prod(BigInt.(Primes.PRIMES[1:1000])); @time factor(x)` shows roughly the same time on this PR vs. master. Do you have test cases where there is a tangible speed-up?

oscardssmith:
It appears that `isprime(x,reps)` is doing a ton of work before the actual test. `isprime(x,50)` and `isprime(x,1)` take exactly the same time. Any idea why this is?

oscardssmith:
After some performance testing, it turns out that the cost of doing all the trial division on a large prime is minuscule compared to the time to do the actual primality tests, so it makes sense to just do all the trial factoring first, and not doing any `isprime` checks. For the `x` you gave, this brings the time down to `.0001` seconds, and for `2*nextprime(x)` (the best-case for the current algorithm), it adds .0001 seconds to a process that already takes .2 seconds.

oscardssmith:
@rfourquet is this good to merge yet? It is a substantial performance improvement for almost all numbers, and is only worse for numbers with exactly 1 small factor (by about .03%).

oscardssmith:
Bumping this. This fixes a long time issue with basically no downside.

oscardssmith:
Bumping. @rfourquet can this merge?

oscardssmith:
@ararslan can you look at this?

rfourquet:
I'm sorry for being unresponsive, I'll look at this over the week-end.

oscardssmith:
Thanks!

rfourquet:
> and is only worse for numbers with exactly 1 small factor (by about .03%).

Did you mean by 3% ?
There are cases with a more pathological behavior yet:
```julia
julia> n = 2*last(Primes.PRIMES)
131042

julia> @btime factor($n) # master
  718.066 ns (3 allocations: 176 bytes)
2 * 65521

julia> @btime factor($n) # PR
  36.485 μs (2 allocations: 160 bytes)
2 * 65521
```
That's roughly 50 times slower. Would there be a way to mitigate this? E.g. by keeping an `isprime` test every X iteration of the loop over `PRIMES` ?

oscardssmith:
Oh, I was testing with bigger numbers (eg `2*(2^127-1)`). For bigger numbers, the prime check is very slow compared to a few trial divisions so this is not worth it, but for smaller numbers, that seems to not be the case. One reasonable solution to this would be to check if the remainder is in `Primes.PRIMES` which can be done in `O(1)`.

rfourquet:
> One reasonable solution to this would be to check if the remainder is in Primes.PRIMES which can be done in O(1)

Yes, but that wouldn't handle numbers like `2 * nextprime(last(Primes.PRIMES) + 1)`.

oscardssmith:
Yeah. I guess the question is where the cutoff is before `isprime` is slower than just doing all the trial division steps.

oscardssmith:
I've just pushed a major performance improvement for numbers smaller than 2^32. Previously this PR would check all primes up to 2^16. Now it checks up to `min(sqrt(n),2^16)`. I'll update with charts once I have them as to what the tipping point is between the time it takes to run `isprime` and do all the required trial division.

oscardssmith:
Here are the performance charts for `Int` and `BigInt` arguments. The TLDR is with my 2 commits today, this PR won't have a major performance impact for numbers <2^14, will be a maximum of about 32x slower for primes around 2^32. For `BigInt` arguments, it will be 50x slower for primes smaller than 2^64, but equal for 2^768, and asymptotically faster afterwards. It is 8x faster at 2^2048, 50x faster for 2^4250, and 1500x faster for 2^44497. 
Int times
BigInt times

Note however that these graphs are the best case for the currently used algorithm. A random number `n` has a roughly 50% chance of having a second largest prime factor greater than `n^.25` (Knuth 4.5.4). Thus for `n>2^64`, the new algorithm will be faster at least 50% of the time when `n` has 2 prime factors `>2^16`. For `n>2^128`, this PR will be better about 3/4ths of the time.

oscardssmith:
For the record, this is the code I used to generate the data
```
function small_factor(x)
	xsqrt = isqrt(x)
	for factor in Primes.PRIMES
		factor > xsqrt && break
		if x%factor==0
			return true
		end
	end
	return false
end

powers = 4:4:1024
times_factor = Float64[]
times_isprime = Float64[]
for power in powers
	p = nextprime(big(2)^power, 2)
	push!(times_factor, @belapsed small_factor(p))
	push!(times_isprime, @belapsed isprime(p))
end
```

oscardssmith:
Here are the figures for random numbers. As you can see, the average case of this PR is much better than what master has. Furthermore, since this PR is also asymptotically better, I think it is worth merging.
Rand Int times
Rand Big times

rfourquet:
Good idea with comparing to the square root! I'm not a specialist and don't have much time right now to dig into benchmarking, so I would be inclined to merge, and possibly iterate later with refined algorithms. Do you have an opinion @ararslan ?

@oscardssmith Would it be worth exploring the following: after a factor is found, we don't call anymore `isprime` here, but say after a number `k` of  iteration over `PRIMES`, if no new prime has been found, we call `isprime` to possibly break the loop? (`k` could be e.g. 500, I'm not sure, I guess it would depend on the relative cost of calling `%` repeatedly for each small prime vs. one call of `isprime`).

oscardssmith:
The thing that makes this complicated is that the tradeoff between `isprime` and `%` varies depending on the size of `n`. The other thing that makes this complicated is that the likelihood of `n` being prime, and of the next factor being close to the most recently found one. Given that this is on average much better than current behavior, I think it makes sense to merge this now, and make iterative improvements from there.

PR #94, opened 2021-02-28
Faster `imsersenne`

oscardssmith:
Use trial division and pohlard's P-1 algorithm to speed up `ismersenne` when false. Based roughly on https://www.mersenne.org/various/math.php. The basic idea is that since almost no mersenne numbers are prime, it's worth doing some quick tests before going to LL.
For example, on  `ismersenne(big(2)^132049-1)` this adds about 2.5 seconds of extra checks (compared to 80 seconds original runtime). However, these checks make the algorithm run in only the 2.5 seconds for the vast majority of outputs.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -459,7 +459,68 @@ function ismersenneprime(M::Integer; check::Bool = true)
             throw(ArgumentError("The argument given is not a valid Mersenne Number (`M = 2^p - 1`)."))
     end
     M < 7 && return M == 3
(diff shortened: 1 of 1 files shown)


oscardssmith:
I've also realized that using `mod_mersenne` in `ll_primecheck` gives a 3.5x speedup for the function in the `true case`.

oscardssmith:
Any ideas why the 32 bit tests are timing out?

PR #98, opened 2021-09-29
Fix ys in Pollard's rho algorithm

mlochbaum:
I was using the Julia implementation of Pollard's rho algorithm as a reference and found the usage of `ys` to be strange: it turns out it doesn't match Brent's [paper](http://maths-people.anu.edu.au/~brent/pub/pub051.html) linked in the source (which does have the unusual style of placing the first statement in a `repeat` block on the same line as that keyword). See commit message for details. The high value of 1900 for `m` might need to be revisited with the corrected algorithm, as setting `ys` too often would have caused the last loop, bounded in size by `m`, to be effectively shortcut.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -411,8 +411,8 @@ function pollardfactors!(n::T, h::AbstractDict{K,Int}) where {T<:Integer,K<:Inte
             k::K = 0
             G = 1
             while k < r && G == 1
+                ys = y
                 for i in 1:(m > (r - k) ? (r - k) : m)
-                    ys = y
                     y = y^2 % n
                     y = (y + c) % n
                     q = (q * (x > y ? x - y : y - x)) % n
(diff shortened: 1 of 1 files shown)


oscardssmith:
What are the consequences of this? Does our version fail more often/ is it slower? If you provide a few tests showing that this works better, I'd be happy to merge.

mlochbaum:
The existing version will fail more often, as the small `while G==1` loop isn't doing its job. So more initial values need to be checked and the algorithm is slower overall.

I don't know Julia very well, so it would probably take me some time to get tests that run with Primes.jl. A test that shows the differences in [my version](https://github.com/mlochbaum/bqn-libs/blob/master/primes.bqn#L191) is to take products of prime pairs that are slightly separated: `prime(1e6)*prime(1e6+1e3)`, `prime(1e6+1)*prime(1e6+1+1e3)`, and so on. I took the total time to factor 200 of these pairs.

I found after filing this PR that a block size of 100 is in fact much faster than 1900, so I've added a commit for that. Here are my timings in BQN for all combinations of patches. The move to 100 eliminates most of the damage from setting `ys` incorrectly, but there's still about a 1% difference that's consistent across changes to the prime numbers tested.

|      | no fix | fix
|------|--------|----
| 1900 | 2.079  | 1.958
|  100 | 1.703  | 1.687

oscardssmith:
This looks good! I'll merge assuming tests pass.

PR #102, opened 2021-12-10
speed up `prime(n)`

oscardssmith:
87x faster for `prime(10^7)`. There are still much better ways of computing this functions, but this is a fairly simple, but massive improvement.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -704,8 +704,21 @@ julia> prime(3)
 
 ```
 """
-prime(::Type{T}, i::Integer) where {T<:Integer} = i < 0 ? throw(DomainError(i)) : nextprime(T(2), i)
-prime(i::Integer) = prime(Int, i)
+prime(::Type{T}, n::Integer) where {T<:Integer} = T(prime(n))
+function prime(n::Integer)
+    n = Int(n)
+    n < 0 && throw(DomainError(i))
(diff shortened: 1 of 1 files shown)


oscardssmith:
The second commit here gives another 40% speedup by using a `BitVector` instead of `Vector{Bool}` `Primes._primemask`. This also gives a roughly 80% speedup in `Primes._primemask` which is considerable.

oscardssmith:
@rfourquet this is ready to review (I just made it a bunch more complicated, sorry). The good news is that this is now 2000x faster than `Master` for `prime(10^7)` (14 seconds vs 7 ms).  With this PR as it stands, `prime(10^11)` can be computed in 10.5 seconds on my computer. This is still not optimal, but is pretty good for only 100 lines of code. The biggest unfortunate thing about this PR in current state is that it adds `SpecialFunctions` as a dependency since it needs the Riemann Zeta function to estimate `prime(n)`. If you think it would be better, I could turn this into a ccall to the MPFR function, which would keep this package as a lighter dependency.

oscardssmith:
@rfourquet one more question: How do you feel about making Primes.jl use multi-threading? Both this PR and https://github.com/JuliaMath/Primes.jl/pull/104 can be sped up more relatively via threading.

Issue #99, opened 2021-11-10
n-th prime is too slow. Consider using primecount as a backend


galanakis:
The n-th prime function of julia is actually really slow.
For example in julia prime(10^7) runs in 14 seconds in my machine.

There is a well known C++ library with a 2-clause BSD license called primecount which can calculate
the prime(10^14) in 2 seconds (of course prime(10^7) is instantaneous.

Mathematica's Prime function is comparable to primecount (it is actually 2 times slower).

So given that, why not use primecount as a backend for the n-th prime function?

I could help with a patch if something like this sounds interesting.

oscardssmith:
It would be much better (in my opinion) to get a fastish Julia implementation. The current implementation is a loop that calls `nextprime`, which is very bad. The decent simple implementation would be a segmented prime sieve that counts up as it finishes the segment. This should be relatively simple to implement and fast.

galanakis:
Thanks for the response. I am interested in implementing the Meissel-Lehmer algorithm. It will take months of part-time work.
Is anyone else working on an implementation of prime-count?

oscardssmith:
https://github.com/JuliaMath/Primes.jl/pull/102 is my 1 hour effort. It computes `prime(10^7)` in `.14` seconds which isn't optimal, but is a lot better.

galanakis:
Yes, it is certainly much better. However primecount is instantaneous even at 10e14, which is really impressive. I am wondering if there is interest in an implementation of the Messel - Lehmer method.

https://www.ams.org/journals/mcom/1996-65-213/S0025-5718-96-00674-6/S0025-5718-96-00674-6.pdf

Is having best of class algorithms in the scope of this project?

oscardssmith:
I'd consider it in scope (although finding a reviewer with sufficient knowledge to approve it might be hard). Also, it seems a little premature to me given that we don't even have a segmented prime sieve yet (or efficient primality checking).

galanakis:
It is true that the algorithm needs a segmented sieve and it is a priority. I could work on that.
Is there a specification or a potential interface for a segmented sieve? Also, it seems that the author of the following project

https://github.com/haampie/FastPrimeSieve.jl

is considering to push it to Primes.jl.

In general what would be the plan for a segmented prime sieve?

oscardssmith:
@haampie would it be reasonable to make `Primes.jl` depend of `FastPrimeSieve`?

galanakis:
[...]

PR #103, opened 2021-12-13
add count_primes

oscardssmith:
Based off of discussion in https://github.com/JuliaMath/Primes.jl/issues/99, this turns the existing prime sieve into a more efficient segmented one. This is about 5x slower than [FastPrimeSieve](https://github.com/haampie/FastPrimeSieve.jl/tree/master/src) (as tested on 10^10), but also only takes about 30 new lines of code. This will be a better foundation for https://github.com/JuliaMath/Primes.jl/pull/102

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -18,17 +18,17 @@ include("factorization.jl")
 #     https://en.wikipedia.org/wiki/Wheel_factorization
 #     http://primesieve.org
 #     Jonathan Sorenson, "An analysis of two prime number sieves", Computer Science Technical Report Vol. 1028, 1991
-const wheel         = [4,  2,  4,  2,  4,  6,  2,  6]
-const wheel_primes  = [7, 11, 13, 17, 19, 23, 29, 31]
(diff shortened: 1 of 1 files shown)


haampie:
Notice that `FastPrimeSieve.pcountprimes` is threaded, which gives more speedup:

```
julia> @time Primes.count_primes(1, 10^10)
 22.230106 seconds (162.77 k allocations: 2.545 MiB)
455052511

julia> @time FastPrimeSieve.countprimes(10^10)
  2.389320 seconds (7 allocations: 277.109 KiB)
455052511

julia> @time FastPrimeSieve.pcountprimes(10^10)
  0.269437 seconds (14.94 k allocations: 4.173 MiB, 0.00% compilation time)
455052511

julia> @time FastPrimeSieve.pcountprimes(10^10, segment_length=128*1024)
  0.195094 seconds (243 allocations: 4.828 MiB)
455052511
```

So it's about a factor 114 slower on my 8 core AMD processor after a bit of tuning

oscardssmith:
I was comparing to single threaded since I also plan to eventually add multithreading to this code. I believe the main place improvement to be made is that currently, the inner loop has a `divrem` to calculate the index that accounts for a lot of the time in the algorithm, and I think it is removable.

haampie:
Yeah, the offset can be saved after the last iteration of each prime number inside every segment

oscardssmith:
Closing this since https://github.com/JuliaMath/Primes.jl/pull/102 is a better path forwards.

PR #104, opened 2021-12-18
add lenstra eliptic curve factoring

oscardssmith:
heavily inspired by @trizen, but refactored a little to make it more reliable.

PR #105, opened 2022-01-12
use `IntegerMathUtils`

oscardssmith:
This PR adds a dependency on https://github.com/JuliaMath/IntegerMathUtils.jl which is a better place to put various integer functions (especially simple ones that wrap GMP). This PR as is doesn't make a ton of sense by itself, but it provides a good foundation for https://github.com/JuliaMath/Primes.jl/pull/104 and https://github.com/JuliaMath/Primes.jl/pull/102 which will want integer nth roots and Integer power checks (checking if `x=a^b` for `a`, `b` integers).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,11 +4,13 @@ version = "0.5.1"
 
 [extras]
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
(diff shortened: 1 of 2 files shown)


oscardssmith:
This should pass CI once https://github.com/JuliaMath/Primes.jl/pull/106 is merged

PR #108, opened 2022-03-29
Allow DataStructures v0.18

ChrisRackauckas:
DataStructures.jl is a core package and v0.18 was released in Aug 17, 2020. This means that pretty much any package made after that won't be compatible with Primes.jl in its latest form, so this shows up as a major ecosystem update blocker. We should really make sure that this kind of repo has CompatHelper.jl on it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,6 +11,6 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 test = ["DataStructures", "IntegerMathUtils", "Test"]
 
 [compat]
-DataStructures = "0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17"
+DataStructures = "0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17, 0.18"
 IntegerMathUtils = "0.1"
 julia = "1"


oscardssmith:
wait, this seems especially silly since the only place we use `DataStructures` is for a [doctest](https://github.com/JuliaMath/Primes.jl/blob/88e348bc2cd6b60902159b63e6e40419e22e22c5/src/Primes.jl#L314).

ChrisRackauckas:
😮‍💨 so make it a test dep?

oscardssmith:
Do you want to make the PR or should I?

ChrisRackauckas:
Can you? I'm going to have to take off and be gone for the rest of the day, but this is pretty necessary since it's causing quite a bit of downstream breakage.

oscardssmith:
This is made slightly more complicated since this package supports Julia 1.0 which doesn't have test decencies (it uses `[extras]` instead. Also, as I understand it, test dependencies still need to be listed in the `[compat]` section (or am I missing something?). As such, I will merge this since it fixes the immediate problem.

ChrisRackauckas:
and Tag?

PR #113, opened 2022-05-02
use cached primes for fast `isprime` of small numbers

oscardssmith:
This reduces the time to 8ns from 250ns for `@btime isprime($(prevprime(2^16)))`. It's slightly slower for small non-primes (numbers with a smallest factor less than 13), but it means that for all small numbers, `isprime` will take very little time (less than 10ns)


oscardssmith:
Closing as obsoleted by https://github.com/JuliaMath/Primes.jl/pull/117

PR #114, opened 2022-05-02
simpler and slightly faster factorization

oscardssmith:
`divrem` is faster than separate `div` and `rem`


oscardssmith:
@rfourquet this is ready for review (and https://github.com/JuliaMath/Primes.jl/pull/115 will be very quick after this is merged). Imo, it's a clear win (slightly faster, slightly cleaner)

PR #118, opened 2022-06-09
Add `eachfactor` iterator

oscardssmith:
followup to https://github.com/JuliaMath/Primes.jl/pull/117. Adds an eachfactor iterator that allows users to get the factors without allocations which can be very nice for lowering overhead of factoring small numbers.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -7,6 +7,7 @@ DocTestSetup = :(using Primes)
 ## Prime factorization
 
 ```@docs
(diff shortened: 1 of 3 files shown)


oscardssmith:
Yeah. I'm planning on merging this pretty soon (and bump the version to 0.5.3). The last thing I need to do is add a few tests.

oscardssmith:
I'll merge this by the end of the week unless there are objections.

Issue #122, opened 2022-06-27
possible number theory functions to add


jmichel7:
Here is a example of two number theory functions made fast by `eachfactor`

```
"""
`prime_residues(n)` the numbers less than `n` and prime to `n`
```julia-repl
julia> [prime_residues(24)]
1-element Vector{Vector{Int64}}:
 [1, 5, 7, 11, 13, 17, 19, 23]
```
"""
function prime_residues(n)
  if n==1 return [0] end
  pp=trues(n) # use a sieve to go fast
  for (p,np) in eachfactor(n)
    pp[p:p:n].=false
  end
  (1:n)[pp]
end

"""
`divisors(n)` the increasing list of divisors of `n`.
```julia-repl
julia> [divisors(24)]
1-element Vector{Vector{Int64}}:
 [1, 2, 3, 4, 6, 8, 12, 24]
```
"""
function divisors(n::Integer)::Vector{Int}
  sort(vec(map(prod,Iterators.product((p.^(0:m) for (p,m) in eachfactor(n))...))))
end

```

oscardssmith:
yeah, I'm not really sure what I want here. On the one hand, these are simple relatively optimal implementations that shouldn't take much maintenance, on the other, they do seem somewhat disconnected to the core of Primes. Does anyone else have strong opinions either way on this one?

wherrera10:
I find I have used this one frequently since Julia 0.7:
    using Primes
 
    function factors(n)
        f = [one(n)]
        for (p, e) in factor(n)
            f = reduce(vcat, [f * p^j for j in 1:e], init = f)
        end
        return length(f) == 1 ? [one(n), n] : sort!(f)
    end
 
which is more or less `divisors` above in a different form. So I think divisors belongs in Primes for me more than totient does.

adrsm108:
I'd also agree that Primes needs a divisors function. I'll throw in my own implementation, which I've tuned a bit for performance since I find myself in need of it surprisingly often. @oscardssmith would you consider a pull request? 

`````julia
"""
    divisors(n::T) -> Vector{T}

Returns all positive divisors of the integer `n`.

For an integer `n` with prime factorization `n = p₁^k₁ ⋯ pₘ^kₘ`, `divisors(n)`
returns a vector of length (k₁ + 1)⋯(kₘ + 1) containing the divisors of `n` in
lexicographical (rather than numerical) order.

```julia
julia> divisors(60)
12-element Vector{Int64}:
  1      # 1
[...]

PR #124, opened 2022-08-05
Divisors function

adrsm108:
This PR introduces the function `divisors`, which takes an integer (or integer factorization), and returns a vector of its positive divisors. 

#### Methods
The function has two methods: `divisors(n::T)::Vector{T} where {T<:Integer}`, and `divisors(f::Factorization{T})::Vector{T} where {T<:Integer}`. The former calls the latter directly. I played with the new `eachfactor` iterator looking for a more efficient technique in the integer case, but was ultimately unsuccessful. Knowing all prime exponents beforehand allows you to preallocate your output array, and it appears the performance benefits of doing so far outweigh the costs of creating a new `Factorization`.

#### Divisor sorting
As I've tried to make clear in its documentation, the output of this function is not numerically sorted. Instead, divisors come in lexicographic order:
$$n=p_1^{k_1} p_2^{k_2} \dots \mapsto (1, p_1, p_1^2, \dots, p_1^{k_1}, p_2, p_1 p_2, p_1^2 p_2, \dots, p_1^{k_1} p_2, p_2^2, \dots ).$$
The algorithm builds the array in this order, and I think returning it without sorting is a suitable design choice because it makes fewer assumptions about the user's purpose. Numerical order, if desired, can be readily obtained by calling `sort!`. Alternatively, `reshape(divisors(p1^k1 * p2^k2 * ... * pm^km), (k1 + 1, k2 + 1, ..., km + 1))` results in an `m`-dimensional array where dimension $i$ corresponds to powers of $p_i$—potentially a useful structure. For yet other uses (Dirichlet convolution comes to mind) order is irrelevant, and speed is the number one priority.

#### Zero
What should `divisors(0)` do? I throw an `ArgumentError`, same as `totient`, but perhaps it would be more convenient to quietly return an empty array?

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -34,4 +34,5 @@ Primes.primesmask
 ```@docs
 Primes.radical
 Primes.totient
+Primes.divisors
 ```
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -1,5 +1,4 @@
 # This file includes code that was formerly a part of Julia. License is MIT: http://julialang.org/license
-
 module Primes
 
 using Base.Iterators: repeated
@@ -9,7 +8,7 @@ using Base: BitSigned
 using Base.Checked: checked_neg
 using IntegerMathUtils
 
-export isprime, primes, primesmask, factor, eachfactor, ismersenneprime, isrieselprime,
+export isprime, primes, primesmask, factor, eachfactor, divisors, ismersenneprime, isrieselprime,
        nextprime, nextprimes, prevprime, prevprimes, prime, prodfactors, radical, totient
 
 include("factorization.jl")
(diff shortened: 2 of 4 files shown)


adrsm108:
Related to #122

oscardssmith:
I think an empty array would be a reasonable behavior for `divisors` (similar to `factor`). I think lexicographic order makes sense, but I would want to hear from others on that topic if possible. One other possibility worth considering: Can the `i`th divisior be determined lazily given the list of factors, and if so, is the right return a lazy object that computes divisiors on the fly?

adrsm108:
I can change the behavior for `divisors(0)`, no problem. 

As for your second question, yeah, that would absolutely be possible. I don't know if a lazy iterator is really the best choice for divisors, though. If you will be using all of them (as is often the case for me), it's considerably more efficient to generate them at once, taking advantage of the fact that every divisor greater than 1 is a prime multiple of some previous divisor. I'd also point out that you'd have to be working with some truly enormous inputs for keeping them all in memory to become an issue, since the number-of-divisors function grows really slowly. (If you're curious like I was, you can look at the highly composite numbers OEIS [A002182](https://oeis.org/A002182) to find that the upper bound on $\sigma_0(n)$ is 1600 (at `n=2095133040`) for an `Int32`, and 161280 (at `n=9200527969062830400`) for an `Int64`).

For cases where you only need a few, perhaps an `nth_divisor` function would be a better alternative? Something like
```jl
nth_divisor(f::Factorization, n::Int) = prod(
        ((p, s),) -> p^(s - 1),
        zip(keys(f), Base._ind2sub(Tuple(values(f) .+ 1), n));
        init = one(keytype(f))
    )
```
You'd still have to take the special cases like 0 and negative factorizations into account, but that's the basic idea.

oscardssmith:
good point. In that case, I think this looks good as is.

PR #127, opened 2023-01-27
suppressed Integer annotation on Factorization; fixed bug in calls of…

jmichel7:
I suppressed the annotations `<:Integer` on the `T` of `Factorization`, allowing `Primes.Factorization` to be used to hold factorizations of other than integer (my goal is for polynomials). I added tests in `runtest.jl` where `T` is `Symbol`.
Along the way, I fixed bugs in the calls to `searchsortedfirst`. In the call `searchsortedfirst(f.pe, p, by=first)` contrary to intuition, the function `by` is applied to `p` so the proper second argument should be a pair. This does not show for `p` integer since then `first(p)==p` but it shows for other types. The fix is to give a pair:
 `searchsortedfirst(f.pe, p=>0, by=first)`

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -966,7 +966,7 @@ end
 Return a vector with the positive divisors of the number whose factorization is `f`. 
 Divisors are sorted lexicographically, rather than numerically.
 """
-function divisors(f::Factorization{T}) where {T<:Integer}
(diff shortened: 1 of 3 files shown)


oscardssmith (approved):

oscardssmith:
Sorry for forgetting about this. Merging once CI passes.

PR #137, opened 2023-09-20
Update Documenter version

oscardssmith:
@fingolfin this might also help with the docs

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -3,4 +3,4 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Primes = "27ebfcd6-29c5-5fa9-bf4b-fb8fc14df3ae"
 
 [compat]
-Documenter = "~0.20"
+Documenter = "~1.0"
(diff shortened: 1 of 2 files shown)


fingolfin:
I don't think that's it. Rather we need to set some doctest metadata. I can submit a PR for this when I am back at my computer; it's a bit too annoying to debug this from my iPad ;-)

oscardssmith:
lets see if the combination works.

fingolfin:
Seems to work. The remaining failure is genuine; the comments in the output were manually added. One could discard them (but that would be a loss), or change the jldoctest into a plain julia block; or perhaps employ a doctest filter...

oscardssmith:
Can you add the PR to make this have a doctest filter (since I don't really understand what that is)

docs/Project.toml
```diff
@@ -3,4 +3,4 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Primes = "27ebfcd6-29c5-5fa9-bf4b-fb8fc14df3ae"
 
 [compat]
-Documenter = "~0.20"
+Documenter = "~1.0"
```
giordano, line 6: You probably wanted to use
```toml
Documenter = "1.0"
```
With `~1.0` you won't be able to upgrade to v1.1.0

PR #142, opened 2023-09-20
Disable failing doctest

fingolfin:
This doctest annotates the output of `divisors(60)`, which is nice, but Documenter doesn't like it. So just turn it into a plain code block without any testing.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -4,6 +4,7 @@ DocMeta.setdocmeta!(Primes, :DocTestSetup, :(using Primes); recursive = true)
 makedocs(
     modules = [Primes],
     sitename = "Primes.jl",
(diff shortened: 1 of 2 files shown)


fingolfin:
@oscardssmith works now!

I had to disable the test which verifies if all docstrings are actually included in the documentation -- they aren't... (and I am torn about that check; I have plenty docstrings in my projects that are not meant for users and hence not included there). Anyway, my point is: one can re-enable this in a future PR. For now, it's more important to restore doc deployment. So let's see whether merging this branch will result in the `gh-pages` branch being updated for the first time in years ;-).

oscardssmith:
Thanks so much for the effort you've put into this!

PR #147, opened 2023-11-15
changed documentation of isprime(n::Integer) -> Bool

fepaul-book:
- shows exactly when probabilistic methods are applied and when not
- mentions the used methods

changed documentation of isprime(x::BigInt, [reps = 25]) -> Bool
- stated where the function is_probably_prime comes from and explains what it does


oscardssmith:
Thanks!

fepaul-book:
cc @ranocha

PR #169, opened 2025-11-04
docs: fix some formatting in the isprime doc string

StefanKarpinski:
This was indented and getting formatted as code. Did a bit of wrapping of long lines while I'm here.


oscardssmith:
thanks!

Issue #159, opened 2024-12-11
Factorisation of "big" numbers


s-celles:
Hello,

I'm trying to factor a quite big number 632459103267572196107100983820469021721602147490918660274601
I tried

```julia
using Primes
factor(632459103267572196107100983820469021721602147490918660274601)
```

but its hanging

https://factordb.com/ https://www.alpertron.com.ar/ECM.HTM or https://sagecell.sagemath.org/ return result in a few second.

How to proceed to achieve such quick proceeding using Julia.

Thanks

s-celles:
Is it possible to proceed directly with package manager mode?

```
(@v1.11) pkg> dev https://github.com/JuliaMath/Primes.jl/pull/104
     Cloning git-repo `https://github.com/JuliaMath/Primes.jl/pull/104`
ERROR: failed to clone from https://github.com/JuliaMath/Primes.jl/pull/104, error: GitError(Code:ERROR, Class:HTTP, request failed with status code: 404)

(@v1.11) pkg> dev https://github.com/oscardssmith/Primes.jl/tree/Lenstra-elliptic-curve
ERROR: rev argument not supported by `develop`; consider using `add` instead
```

oscardssmith:
you need `]add https://github.com/oscardssmith/Primes.jl.git#Lenstra-elliptic-curve`

oscardssmith:
ah. Looks like the branch isn't currently working. that's somewhat unfortunate.

oscardssmith:
The branch is now working. However, ECM isn't up to the task here Looks like to solve this quickly we would need a quadratic sieve implementation. (for comparison, your number is has both factors at roughly 2^100)

```
julia> @time factor(nextprime(big(2)^62)*nextprime(big(2)^102, 2))
  8.508423 seconds (107.79 M allocations: 2.661 GiB, 11.26% gc time)
4611686018427388039 * 5070602400912917605986812821829

julia> @time factor(nextprime(big(2)^72)*nextprime(big(2)^102, 2))
 61.739692 seconds (788.39 M allocations: 19.518 GiB, 11.07% gc time)
4722366482869645213711 * 5070602400912917605986812821829
```

s-celles:
It's hanging.
I wonder if we shouldn't have a factor function which is using multiple dispatch according to factorization method that user wants to use. I think it will make the code clearer.

Here is an example (LLM code)

```julia
using Primes
using Combinatorics

# Abstract type for different factorization strategies
[...]

PR #174, opened 2026-03-12
feat: Elliptic Curve Method (ECM) for integer factorization

s-celles:
- ecm_factor(n, B1, num_curves) — attempts to find a non-trivial factor of n using ECM Stage 1.
 - Two optimized code paths, selected automatically by dispatch:
   - BigInt: in-place GMP arithmetic via ccall (ECMBuffers with 12 preallocated BigInts, zero allocation in the hot loop).
   - Generic T<:Integer: works for UInt128, Int128, and BitIntegers.jl types (UInt256, UInt512, etc.). Modular multiplication uses widen(T) to stay in fixed-width LLVM integers, avoiding BigInt heap allocation.
 - Overflow-safe modular arithmetic (_addmod, _submod, _mulmod) that handles unsigned types near typemax correctly.
 - Shared helpers (_ecm_prime_powers, _ecm_suyama) factored out to avoid duplication between the two paths

Helps #159 
Part of #173 with some BitIntegers.jl types support


src/ecm.jl
```diff
@@ -0,0 +1,353 @@
...
+@inline function _submod(a::T, b::T, n::T) where {T<:Integer}
+    a >= b ? a - b : n - (b - a)
+end
+
+# Modular multiplication: (a * b) mod n.
+# Uses widen(T) to avoid overflow for fixed-width integers.
+# For BigInt, widen returns BigInt itself, so this is simply mod(a * b, n).
+@inline function _mulmod(a::T, b::T, n::T) where {T<:Integer}
+    W = widen(T)
+    T(mod(W(a) * W(b), W(n)))
```
oscardssmith: Can you write this without the `widen`? (under the assertion that `n^2<typemax(T)`). That's what we do for pollard and it seems to work pretty well (the advantage is that `widen(Int128)` is `BigInt` (and in general widen/narrow will be somewhat expensive).

oscardssmith:
Eventually, it would be good to have the 2 stage version of ECM, but the single stage version should already be an improvement.

s-celles:
This PR is a concrete motivation for #176. `_ecm_prime_powers` currently calls `primes(B1)` which allocates a full vector upfront:

```julia
function _ecm_prime_powers(B1::Int)
    result = Int[]
    for p in primes(B1)   # allocates all primes ≤ B1
        pk = p
        while pk * p <= B1
            pk *= p
        end
        push!(result, pk)
    end
    result
end
```

With `primes_from` from #176, the inner loop becomes lazy — no upfront allocation:

```julia
function _ecm_prime_powers(B1::Int)
    result = Int[]
    for p in Iterators.takewhile(<=(B1), primes_from(2))
        pk = p
        while pk * p <= B1
            pk *= p
        end
        push!(result, pk)
    end
    result
end
```

For typical ECM usage with large `B1`, this would reduce memory pressure in the Stage 1 setup. And if `primes_from` is eventually backed by a segmented sieve (as discussed in the planned refactor), the benefit would come for free — without any change to the ECM code itself.

oscardssmith:
IMO this is a fairly weak motivation as a B1 of 11_000 requires very little memory. I do agree that it is a good idea though.

PR #184, opened 2026-07-22
Add `sieve_bound` to `SegmentedSieve`, Make `SegmentedSieve` generic, and narrow-window `eachprime`

oscardssmith:
This is a 4 small but nice improvements to the newly added `SegmentedSieve` and `eachprime`.

1. Add `sieve_bound` to `SegmentedSieve` for B-rough sieving (survivors have no prime factor <= B); default isqrt(hi) preserves the exact prime sieve.
2. Pack `stride_primes` into `Vector{Int64}`, stored as (`256*(p/30)+wheel_index`). This allows sieving by primes up to 
3. Parametrize `SegmentedSieve{T}`/EachPrime{T} over the integer type to allow `Int128`/`BigInt heights (note that this will only work for sieving up to a .  Per-bit offsets stay Int; T arithmetic is confined to O(pi(bound))/window sites.
4. eachprime uses a narrow-window strategy: When `hi-lo < sqrt(hi)/150` it sieves only by primes `p < hi-lo` and uses `isprime` to ensure it only returns primes. This is much faster and lowers memory for `eachprime` from `O(sqrt(hi) + hi-lo)` to `O(min(sqrt(hi), (hi-lo) * ln(ln(hi)) ))`.

Benchmarks:
```
julia> @time sum(eachprime(10^18, 10^18+10^5));
  1.606364 seconds (336 allocations: 1.317 GiB)   # before
  0.004651 seconds (123 allocations: 323.164 KiB) # after

julia> @time sum(eachprime(10^18, 10^18+10^6))
  1.642741 seconds (350 allocations: 1.317 GiB)   # before 
  0.042451 seconds (152 allocations: 2.750 MiB)   # after

julia> @time sum(eachprime(10^18, 10^18+10^7));
  1.650434 seconds (358 allocations: 1.322 GiB)   # before
  1.086680 seconds (304 allocations: 411.965 MiB) # after
```

Coauthored by claude.

### DIFF
--- a/src/Primes.jl
+++ b/src/Primes.jl
@@ -84,7 +84,10 @@ false
 ```
 """
 function isprime(n::T, reps::Integer=25) where {T<:Integer}
-    n ≤ typemax(Int64) && return isprime(Int64(n))
+    n < 2 && return false
+    # Everything below 2^64 uses the deterministic UInt64 path (hardware UInt128 arithmetic); for wider
+    # T (e.g. Int128) staying in T would make `widemul` widen to BigInt and allocate.
+    n ≤ typemax(UInt64) && return isprime(UInt64(n))
     # GMP-style probable-prime test in the operand's own arithmetic
     # BPSW, then reps - 24 extra Miller-Rabin rounds so
     # the error bound tracks GMP's for the same reps.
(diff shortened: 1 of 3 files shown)


oscardssmith:
@s-celles I think this fixes almost all usecases https://github.com/JuliaMath/Primes.jl/pull/178 used to be needed for.

